PR #1, opened 2017-01-30
APB-568: Enable scoverage and ScalaStyle

frankoid:


frankoid:
eyeballed by @tev1622

PR #3, opened 2017-02-01
[apb-569][ph] added mock subscription details page

PaulHodgson:


frankoid:
It's a small thing but please can we use upper case JIRA refs, at least until we've confirmed that cross referencing from JIRA works with lower case ones?

PR #7, opened 2017-02-03
[APB-488][ph] first stab at auth on check agency status

PaulHodgson:
first stab with some commented code and tests. This is deliberate so that we can parallelise development more quickly.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/auth/MtdSubscriberRegime.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/auth/MtdSubscriberRegime.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2017 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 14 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/auth/MtdSubscriberRegime.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package uk.gov.hmrc.agentsubscriptionfrontend.auth
+
+import uk.gov.hmrc.play.frontend.auth
+import uk.gov.hmrc.play.frontend.auth.TaxRegime
+import uk.gov.hmrc.play.frontend.auth.connectors.domain.Accounts
+
+object MtdSubscriberRegime extends TaxRegime {
+  override def isAuthorised(accounts: Accounts) = accounts.sa.isDefined
```
frankoid: We're expecting the agent to use a new GG cred so they won't have an IR-SA enrolment (which is what accounts.sa is mapped from).

I think we probably want isAuthorised to always return true (or something else that provides equivalent behaviour, i.e. you have to be logged in but we don't perform any further checks)

PaulHodgson: Ah okay. Overkill for this task in any case, will correct.

PaulHodgson: Pushed a change...

PR #10, opened 2017-02-08
[FD][APB-493] Add test for error handling when getting affinity group

frankoid:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -31,17 +31,18 @@ import uk.gov.hmrc.play.http.HeaderCarrier
...
+class SubscriptionController @Inject() (override val messagesApi: MessagesApi, override val authConnector: AuthConnector) (implicit appConfig: AppConfig)
   extends FrontendController with I18nSupport with Actions {
 
-  override protected def authConnector: AuthConnector = FrontendAuthConnector
-
-  val showCheckAgencyStatus: Action[AnyContent] = AuthorisedFor(NoOpRegime, GGConfidence).async { implicit authContext: AuthContext => implicit request =>
-    ensureAffinityGroupIsAgent {
-      Ok(html.subscribe())
-    }
+  val showCheckAgencyStatusBody: (AuthContext) => (Request[AnyContent]) => Future[Result] = {
```
frankoid: The reason for breaking out `showCheckAgencyStatusBody` like this is to make unit testing easier.

`showCheckAgencyStatusBody` can be tested without auth mocking to make the `AuthorisedFor` check pass.

This also avoids the need for the Play app to be started in `SubscriptionControllerSpec` (which is fairly slow). This is because for the `AuthorisedFor` check to pass we'd need to add `.withSession(sessionKeysForMockAuth: _*)` to the `FakeRequest` and `.withSession()` needs a running app because it uses Play's `Crypto` object to sign the session cookie and in turn `Crypto` requires a running app.

bravesirrobin81: Nice. Should perhaps be package-private visibility, though.

project/FrontendBuild.scala
```diff
@@ -27,7 +27,8 @@ object FrontendBuild extends Build with MicroService {
     "org.pegdown" % "pegdown" % "1.6.0" % scope,
     "org.jsoup" % "jsoup" % "1.8.1" % scope,
     "com.typesafe.play" %% "play-test" % PlayVersion.current % scope,
-    "org.scalatestplus.play" %% "scalatestplus-play" % "1.5.1" % scope
+    "org.scalatestplus.play" %% "scalatestplus-play" % "1.5.1" % scope,
+    "org.mockito" % "mockito-core" % "2.7.4" % scope
```
bravesirrobin81: Mockito 2, how daring!

frankoid: I know. Ben upgraded agent-paye to 2.5.0 during the Play upgrade and since this is a new service I thought I'd try the latest version.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
frankoid: > Should perhaps be package-private visibility

agreed, I've pushed a change to make it package-private

PR #12, opened 2017-02-14
Apb 498

PaulHodgson:


app/uk/gov/hmrc/agentsubscriptionfrontend/models/DesBusinessPartnerRecordApiResponse.scala
```diff
@@ -16,8 +16,6 @@
```
PaulHodgson: I think this should be removed.

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnector.scala
```diff
@@ -0,0 +1,45 @@
...
+import scala.concurrent.Future
+
+@Singleton
+class AgentSubscriptionConnector @Inject() (@Named("agent-subscription-baseUrl") baseUrl: URL, httpGet: HttpGet) {
+
+  def getRegistration(utr: String, postcode: String)(implicit hc: HeaderCarrier): Future[Option[Registration]] = {
+    val url = getRegistrationUrlFor(utr, postcode)
+    httpGet.GET[HttpResponse](url).map { response: HttpResponse =>
+      response.status match {
+        case 200 => Some(new Registration)
```
cokleisli: Would be good to change the 200 to the Ok const.

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/HttpDesBusinessPartnerRecordApiConnector.scala
```diff
@@ -1,68 +0,0 @@
...
-}
-
-@Singleton
-class HttpDesBusinessPartnerRecordApiConnector @Inject()(@Named("des-baseUrl") desBaseUrl: URL,
-                                                         @Named("des.authorization-token") authorizationToken: String,
-                                                         @Named("des.environment") environment: String,
-                                                         httpGet: HttpGet) extends DesBusinessPartnerRecordApiConnector {
-
-  def getBusinessPartnerRecord(utr: String)(implicit hc: HeaderCarrier): Future[DesBusinessPartnerRecordApiResponse] = {
-    val url: String = bprUrlFor(utr)
```
cokleisli: I'm not a big fan of explicit type annotations. Less code the better imho. But if the team's policy is to explicitly annotate then that is fine.

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnector.scala
PaulHodgson: done

it/uk/gov/hmrc/agentsubscriptionfrontend/stubs/AgentSubscriptionStub.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.github.tomakehurst.wiremock.client.WireMock._
+import uk.gov.hmrc.play.encoding.UriPathEncoding.encodePathSegment
+
+object AgentSubscriptionStub {
+
+  def withMatchingUtrAndPostcode(utr: String, postcode: String): Unit = {
+    stubFor(get(urlEqualTo(s"/agent-subscription/registration/${encodePathSegment(utr)}/postcode/${encodePathSegment(postcode)}"))
+      .willReturn(
+        aResponse()
+          .withStatus(200)))
```
cokleisli: Same again for these integer response codes - convert them to the Play HTTP response constants.

PaulHodgson: done

frankoid:
I actually quite like it that tests (WireMock stubs etc) use integer status codes instead of Play constants, I think it's a bit clearer and it also catches a certain kind of mistake (if you get your status code names/numbers mixed and use an integer in the test and a constant in the actual implementation then the test will fail, whereas if you use constants for both you'll probably make the same mistake in both places).

I don't feel strongly about it though, we can use constants in tests if others greatly prefer them.

bravesirrobin81:
I'd echo this: we should use separate constants for tests vs production code.

PR #13, opened 2017-02-15
[FD][APB-498] Fix UTR validation

frankoid:
Previously no validation error message would be displayed when an invalid UTR was entered.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/views/subscribe.scala.html
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/views/subscribe.scala.html
@@ -1,10 +1,9 @@
 @import uk.gov.hmrc.agentsubscriptionfrontend.config.AppConfig
 @import uk.gov.hmrc.play.views.html.helpers.form
+@import uk.gov.hmrc.play.views.html.helpers.input
 
 @import uk.gov.hmrc.agentsubscriptionfrontend.controllers.routes
 @import uk.gov.hmrc.agentsubscriptionfrontend.controllers.KnownFacts
-
-@import helper._
 
 @(knownFactsForm : Form[KnownFacts])(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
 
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/views/subscribe.scala.html
```diff
@@ -15,12 +14,23 @@
   @form(
     action = routes.SubscriptionController.submitKnownFacts,
     'class -> "form js-form") {
-    <fieldset class="form-field-group">
-      @helper.inputText(knownFactsForm("utr"))
-      @helper.inputText(knownFactsForm("postcode"))
+
```
bravesirrobin81: What was the reason for moving away from the helpers?

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+    "give \"error.required\" error when it only contains a space" in {
+      bind(" ").left.value should contain only FormError("testKey", "error.required")
+    }
+
+    "give \"error.utr.invalid\" error" when {
+      "it has more than 10 digits" in {
+        bind("12345678901") should matchPattern { case Left(List(FormError("testKey", List("error.utr.invalid"), _))) => }
+      }
+
+      "it has less than 10 digits" in {
```
bravesirrobin81: "fewer"

app/uk/gov/hmrc/agentsubscriptionfrontend/views/subscribe.scala.html
frankoid: I lifted the UI code from agent-client-authorisation-frontend and it uses (HMRC) play-ui helpers instead of using Play's helpers directly. Presumably in order to generate HTML that matches the CSS but I'm not sure.

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
frankoid: I'll change it

PR #17, opened 2017-02-21
Apb 598 Walking skeleton and authorisation refactor

bravesirrobin81:


frankoid:
`sbt clean test it:test` fails with a compile error because `SessionKeysForTesting` is defined in `it` but used in `test` by `AuthActionSpec`. I think there are 2 options to fix, either move `AuthActionSpec` to `it` (one way of looking at it is that using `WithFakeApplication` makes something an integration test) or add a `testcommon` directory like we have in `agent-client-authorisation`.

bravesirrobin81:
Creating a test-common to share a single string constant seems overkill to me. I've duplicated the class. Oh, and fixed the innumerable other reasons why I shouldn't have created this PR.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -0,0 +1,48 @@
...
+import uk.gov.hmrc.play.frontend.auth.Actions
+import uk.gov.hmrc.play.frontend.auth.connectors.AuthConnector
+import uk.gov.hmrc.play.frontend.controller.FrontendController
+
+class StartController @Inject()(override val messagesApi: MessagesApi,
+                                override val authConnector: AuthConnector)
+                               (implicit appConfig: AppConfig)
+    extends FrontendController with I18nSupport with Actions {
+
+  val root: Action[AnyContent] = Action { implicit request =>
```
frankoid: I can't find any tests for `root` or `start`

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -0,0 +1,127 @@
...
+    )(KnownFacts.apply)(KnownFacts.unapply)
+  )
+
+  private[controllers] val showCheckAgencyStatusBody: AuthContext => (Request[AnyContent]) => Future[Result] = {
+    implicit authContext =>
+      implicit request =>
+          Future successful Ok(html.check_agency_status(knownFactsForm))
+  }
+
+  val showAlreadySubscribed: Action[AnyContent] = AuthorisedWithSubscribingAgent { implicit authContext => implicit request =>
```
frankoid: I think this & some of the other actions in this controller are also missing tests

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
```diff
@@ -0,0 +1,61 @@
...
+  "start" should {
+    "not require authentication" in {
+      AuthStub.userIsNotAuthenticated()
+
+      val result = await(controller.start(FakeRequest()))
+
+      status(result) shouldBe 200
+    }
+
+    "be available at /agent-subscription/start" in {
```
frankoid: I don't think this test actually tests that start is available at this particular URL, because it calls the controller method directly instead of using `route()`.

So far we have been leaving tests of the URLs for end-to-end testing so I suggest just changing the description rather than the implementation of the test.

frankoid: We did originally start using `route()` but were unable to pass session keys into it.

Since then I came accross something that might allow us to pass session keys when using `route()` so if you're interested we can look into that.

Not directly relevant to this test except that it would be a bit inconsistent to use `route()` here but not in the tests for authenticated endpoints that need session keys.

it/uk/gov/hmrc/agentsubscriptionfrontend/support/EndpointBehaviours.scala
```diff
@@ -0,0 +1,34 @@
...
+import play.api.test.FakeRequest
+import uk.gov.hmrc.agentsubscriptionfrontend.controllers.routes
+import uk.gov.hmrc.agentsubscriptionfrontend.stubs.AuthStub
+import uk.gov.hmrc.agentsubscriptionfrontend.support.SampleUsers.individual
+import uk.gov.hmrc.play.test.UnitSpec
+
+trait EndpointBehaviours {
+  me: UnitSpec with WireMockSupport =>
+
+  protected def anAgentAffinityGroupOnlyEndpoint(doRequest: FakeRequest[AnyContentAsEmpty.type] => Result) = {
```
frankoid: IntelliJ wants this to have a return type specified because it is a protected (not private) method, would be nice to add this (`: Unit`) just to keep this file free of warnings.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
bravesirrobin81: Should just be this one and the hasOtherEnrolments endpoint. I haven't written the tests yet, as they will come with the meat of this story.

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
frankoid: I'm going to have a quick look to see if we can pass session keys into `route()` - if it's easy it would be nice for our tests to be able to use it (e.g. to cover URLs)

bravesirrobin81: I would question the utility of testing the URLs anyway---as long as we're using the routes object internally, and don't support jumping into the journey midway I don't think it matters what the URLs are.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
frankoid: OK, will merge this but I'm not really comfortable with merging code without tests to master, I think it's best not to add the code until it has tests. In other words, do TDD or if you insist on not doing TDD then make the code/commits look like they were TDDed :)

PR #18, opened 2017-02-22
Apb 598 Check if agent is already subscribed, or has other enrolments

bravesirrobin81:


it/uk/gov/hmrc/agentsubscriptionfrontend/stubs/AuthStub.scala
```diff
@@ -71,6 +71,42 @@ object AuthStub {
...
   }
 
+  def isSubscribedToMtd(user: SampleUser): Unit = {
+    stubFor(get(urlEqualTo(user.enrolmentsLink))
+        .willReturn(
+          aResponse()
+              .withStatus(200)
+              .withBody(
+                s"""
+                   |[{"key":"HMRC-AS-AGENT","identifiers":[{"key":"Arn","value":"ARN0001"}],"state":"Activated"}]
```
frankoid: where does the `Arn` value for identifier key come from? Have you seen the service definition (Q Pack) for HMRC-AS-AGENT?

bravesirrobin81: That's a very good question--I think it was originally a guess as we didn't have the q-pack. It works integrated with auth, which is probably why I didn't question it further---the changes to gg-stubs didn't include an identifier key for the enrolment at all.

bravesirrobin81: And the reason it works is that we check only the presence of the enrolment, and not any identifiers.

frankoid: Perhaps we can remove it until we know what the key will be? I'm looking at the code now to see whether it is actually needed.

frankoid: Seems fine without it (tests pass), I've pushed a commit that removes it

frankoid:
Looks like we don't need the "has other enrolments" check (requirements change) but we do still want the "already subscribed" check so I'm still going to merge this to get the refactorings and already subscribed check if you think that's a good idea @bravesirrobin81. Once the requirements change is confirmed we can remove the "has other enrolments" check.

bravesirrobin81:
Fine by me.

PR #22, opened 2017-03-03
Apb 628 confirm your agency

bravesirrobin81:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -88,7 +88,9 @@ class CheckAgencyController @Inject()
                                             (implicit authContext: AuthContext, request: Request[AnyContent]): Future[Result] = {
     agentSubscriptionConnector.getRegistration(knownFacts.utr, knownFacts.postcode) map { maybeRegistration: Option[Registration] =>
       maybeRegistration match {
-        case Some(_) => Redirect(routes.CheckAgencyController.showConfirmYourAgency())
+        case Some(Registration(Some(name))) => Redirect(routes.CheckAgencyController.showConfirmYourAgency())
+                          .flashing("agencyName" -> name, "knownFactsPostcode" -> knownFacts.postcode, "utr" -> knownFacts.utr)
```
frankoid: The story says `Please use the Name and postcode data from this api
https://confluence.tools.tax.service.gov.uk/display/AG/Registration` but this gets the postcode from the user input, not the API.

I think that the code is probably OK as it is, the only possible difference is the casing and spacing of the postcode displayed, but please can you confirm with JP and assuming he's OK with it update JIRA to match the code?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -88,7 +88,9 @@ class CheckAgencyController @Inject()
...
+        case Some(_) => InternalServerError("No organisation name for agency found in ETMP.")
```
frankoid: What does the UI look like when this happens? Do you see the generic MDTP error page or an unstyled page with the `No organisation name for agency found in ETMP.` message? If the latter I think that we should be showing the generic MDTP error page which could be achieved by throwing an exception here.

Also I think we should be logging to Kibana when this happens (including UTR in the message) - could probably acheive this either with a Logger.warn or by throwing an exception.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
bravesirrobin81: Spoken to JP and JIRA updated.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
bravesirrobin81: The latter. I'll update the code to throw an exception.

PR #26, opened 2017-03-08
[FD] Use keystore for transient storage instead of GET parameters and flash scope

frankoid:
Using GET parameters for UTR & Postcode is insecure because it results in them being included in web server logs.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SessionDataMissing.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SessionDataMissing.scala
@@ -22,7 +22,7 @@ import play.api.mvc.{Result, Results}
 trait SessionDataMissing {
   this: Results =>
 
-  def showSessionDataMissing(): Result = {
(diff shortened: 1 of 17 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SessionDataMissing.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package uk.gov.hmrc.agentsubscriptionfrontend.controllers
+
+import play.api.Logger
+import play.api.mvc.{Result, Results}
+
+trait SessionDataMissing {
+  this: Results =>
+
+  def showSessionDataMissing(): Result = {
```
bravesirrobin81: I don't think the name of this method reflects what it does. How about 'redirectIfSessionMissing'?

frankoid: changed

PR #28, opened 2017-03-09
Apb 622 subscription validation

bravesirrobin81:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
```diff
@@ -47,7 +47,18 @@ package object controllers {
       }
     }
 
+    private val nonEmptyTelephoneNumber: Constraint[String] = Constraint[String] { fieldValue: String =>
+      Constraints.nonEmpty(fieldValue) match {
+        case i: Invalid => i
+        case Valid => (fieldValue.length, fieldValue.replaceAll("[^0-9]", "").length) match {
+          case (length, digitCount) if length > 24 || digitCount < 10 => Invalid(ValidationError("error.telephone.invalid"))
```
PaulHodgson: The code makes sense given: https://confluence.tools.tax.service.gov.uk/display/AG/Field+Validation. Not clear why those literals are used when you come to the code blind though. How to make it a bit clearer? Use named constants rather than literal 24 and 10?

PR #36, opened 2017-03-15
[RG,MK][APB-687] Add Modulus11Check validation for UTR fields.

frgomes-hmrc:


frgomes-hmrc:
Work in progress. We still need to create acceptance tests according to story
https://jira.tools.tax.service.gov.uk/browse/APB-687

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
```diff
@@ -45,15 +45,15 @@ class FieldMappingsSpec extends UnitSpec with EitherValues {
 
     "give \"error.utr.invalid\" error" when {
```
bravesirrobin81: I'd expect to see a case here that explicitly tests for checksum failure.

bravesirrobin81:
There will probably need to be some changes to agent-stubs and the E2E tests as some of our test data may not pass this validation now. Are you looking at that?

PR #42, opened 2017-03-29
[BW][APB-734] - Basic copy updates added to enrolment and affinity gr…

benwatsonuk:
Basic copy updates added to enrolment and affinity group pages. URL to Government Gateway is currently an absolute URL.


frankoid:
Looks like a couple of tests (`sbt test it:test`) need updating for the new copy - I'll update them and then merge this.

Also as discussed the gateway URL should be a setting so it can be different in the QA env, I/someone will do that in a separate PR.

PR #48, opened 2017-04-25
[AO][APB-772] add dependency to agent-mtd-identifiers lib

arturopala:
- add dependency to agent-mtd-identifiers lib
- validate using new `Utr.isValid(...)`
- switch from String to `Utr`


frankoid:
We probably only need one `"give \"error.utr.invalid\" error"` case in `FieldMappingsSpec` now because the tests in agent-mtd-identifiers should cover all the different reasons why UTRs can be invalid - we just need to check that the validation is happening.

I think we made a similar change to agent-mapping or agent-mapping-frontend when we made it use the library.

PR #50, opened 2017-05-04
APB-724 Added further validation requirements for DES

ZDevelop94:


it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -251,14 +263,24 @@ class SubscriptionControllerISpec extends BaseControllerISpec with SessionDataMi
...
 
-        val result = await(controller.submitSubscriptionDetails(subscriptionDetailsRequest("telephone", Seq("telephone" -> "12345"))))
+        val result = await(controller.submitSubscriptionDetails(subscriptionDetailsRequest("telephone", Seq("telephone" -> "nine"))))
 
         status(result) shouldBe 200
-        checkHtmlResultWithBodyText(result, "Add your agency information")
+        checkHtmlResultWithBodyText(result, "Add your agency information", "Please enter a valid telephone number")
+      }
+
+      "telephone is invalid with numbers and words" in {
```
frankoid: It would be better to just have one `"telephone is invalid"` test at the integration level (i.e. in this file) and then have tests for all the different reasons a telephone number might be invalid at the unit level (e.g. in FieldMappingsSpec). Have a look and see if this is possible.

frankoid: same for the name field

conf/messages
```diff
@@ -2,4 +2,6 @@ error.postcode.invalid = Please enter a valid postcode
 error.utr.invalid = Please enter a valid UTR
 error.telephone.invalid = Please enter a valid telephone number
 error.no.ampersand = Agency name cannot contain "&"
-
+error.name.invalid = "This field is mandatory and is limited to 40 alphanumeric characters (A-Z, a-z 0-9 \\-,.)"
```
frankoid: I don't think these 3 messages need to mention "This field is mandatory" because if you don't enter a value you don't see one of these three messages, you see the "This field is required"

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
```diff
@@ -53,9 +60,43 @@ package object controllers {
...
+        }
+      }
+    }
+
     def utr: Mapping[Utr] = nonEmptyText.transform[Utr](Utr.apply,_.value).verifying("error.utr.invalid", utr => Utr.isValid(utr.value))
     def postcode: Mapping[String] = text verifying nonEmptyPostcode
-    def telephoneNumber: Mapping[String] = text verifying nonEmptyTelephoneNumber
-    def agencyName: Mapping[String] = nonEmptyText(maxLength = 40) verifying noAmpersand
+    def telephoneNumber: Mapping[String] = text verifying TelephoneNumberConstraint
+    def agencyName: Mapping[String] = text verifying noAmpersand verifying agencyNameConstraint
```
frankoid: I think we can avoid having so many different custom constraints here.

Step 1 - this doesn't change any behaviour (I think) - remove address23Constraint because it is the same as addressConstraint (isn't it?)
```
    def addressLine1: Mapping[String] = text verifying addressConstraint
    def addressLine23: Mapping[Option[String]] = optional(text verifying addressConstraint)
```

Step 2 - this does change behaviour because the message you'd get for input too long would be different but I think this is a positive behaviour change:
```
    def agencyName: Mapping[String] = text(maxLength = 40) verifying noAmpersand verifying desPlainTextConstraint
    def addressLine1: Mapping[String] = text(maxLength = 35) verifying desPlainTextConstraint
    def addressLine23: Mapping[Option[String]] = optional(text(maxLength = 35) verifying desPlainTextConstraint)
```

desPlainTextConstraint would validate against the regex `"^[A-Za-z0-9 \\-,.&'\\/]*$"`, i.e. the same regex as address and agency name except with the length part (`{1,35}` or `{1,40}`) changed to `*`.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
```diff
@@ -40,11 +45,13 @@ package object controllers {
       }
     }
 
-    private val nonEmptyTelephoneNumber: Constraint[String] = Constraint[String] { fieldValue: String =>
+    private val TelephoneNumberConstraint: Constraint[String] = Constraint[String] { fieldValue: String =>
```
frankoid: inconsistent casing (capital T but other similar things use lower case first letter)

conf/messages
```diff
@@ -2,4 +2,5 @@ error.postcode.invalid = Please enter a valid postcode
 error.utr.invalid = Please enter a valid UTR
 error.telephone.invalid = Please enter a valid telephone number
 error.no.ampersand = Agency name cannot contain "&"
-
+error.des.text.invalid = This field is limited to alphanumeric characters (A-Z, a-z, 0-9) and the following characters \\-,.)/
+error.maxLength = Maximum amount of characters exceeded
```
frankoid: Do we need this custom error.maxLength message? The (default message)[https://github.com/playframework/playframework/blob/2.5.12/framework/src/play/src/main/resources/messages.default#L12] has the advantage that it tells you what the maximum length is.

ZDevelop94: True we do not

frankoid:
I think we have more testing than we need at the integration level in SubscriptionControllerISpec - we have tests there that duplicate testing done in FieldMappingsSpec.

The integration test only needs to test that the FieldMapping is correctly integrated into the SubscriptionController - so for each field that uses one of our custom FieldMappings we only need one "invalid" test in SubscriptionControllerISpec.

Testing of all the different validation scenarios only needs to be done at the unit level in FieldMappingsSpec.

Please can you review SubscriptionControllerISpec for tests that duplicate tests in FieldMappingsSpec and remove them. I've done this already for the telephone field as an example.

frankoid:
Also the tests in SubscriptionControllerISpec that check that a validation failure occurs without checking for the validation failure message seem unreliable to me - they might pass if a different validation failure occurs e.g. a failure related to a different field.

Hopefully removing the duplicate tests will get rid of most of the tests that don't check the validation message though.

ZDevelop94:
I thought the same thing but i didn't want to remove a test someone else did and was still valid.

frankoid:
yes it's good to discuss this kind of thing before going in and deleting loads of code!

I think the email ones should be left in (for now at least) but the ones related to our custom FieldMappings should go - but before deleting each test from SubscriptionControllerISpec make sure it is duplicated in FieldMappingsSpec. And make sure you leave one test in per field so that we have a test that will fail if SubscriptionController's form is changed to not use the FieldMapping.

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -211,16 +192,6 @@ class SubscriptionControllerISpec extends BaseControllerISpec with SessionDataMi
         checkHtmlResultWithBodyText(result, "Add your agency information")
       }
 
-      "email has no text in the local part" in {
```
frankoid: did you mean to remove this test? I saw that the email tests were removed and then added back in again

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
```diff
@@ -170,35 +173,256 @@ class FieldMappingsSpec extends UnitSpec with EitherValues {
         bind("    ").left.value should contain only FormError("testKey", "error.required")
       }
 
-      "fewer than 10 digits" in {
-        shouldRejectFieldValueAsInvalid("12345         ")
-
-      }
       "more than 24 characters" in {
         shouldRejectFieldValueAsInvalid("999999999999999999999999999999999")
```
frankoid: The error you get when the telephone number is too long isn't very helpful because it doesn't tell you that the reason your input was rejected was that it was too long. Perhaps it should be rejected with an `error.maxLength` in that case, just like agencyName and the address fields?

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
```diff
@@ -170,35 +173,256 @@ class FieldMappingsSpec extends UnitSpec with EitherValues {
...
+        shouldAcceptFieldValue("My Agency address")
+      }
+
+      "field is empty" in {
+        shouldAcceptFieldValue("")
+      }
+    }
+  }
+
+  "name bind" should {
```
frankoid: `"agencyName bind"`?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
```diff
@@ -40,11 +42,13 @@ package object controllers {
       }
     }
 
-    private val nonEmptyTelephoneNumber: Constraint[String] = Constraint[String] { fieldValue: String =>
```
frankoid: the old name, `nonEmptyTelephoneNumber`, seems more descriptive than `telephoneNumberConstraint`

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
ZDevelop94: Yes, this test is still there.

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
ZDevelop94: I called the test "name bind" because that is what the field is called, not agencyName. Shouldn't the tests reflect what we are actually testing?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
ZDevelop94: Shouldn't we be consistent with the other variables? If I change this shouldn't we change all of them?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
frankoid: oh yes it has just moved

test/uk/gov/hmrc/agentsubscriptionfrontend/controllers/FieldMappingsSpec.scala
frankoid: what is being bound in this test? `name` or `agencyName`?

ZDevelop94: yeah AgencyName

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
ZDevelop94: ?

frankoid: Consistency is good, descriptive names are also good.
Have you implemented (this)[https://github.com/hmrc/agent-subscription-frontend/pull/50#discussion_r115237213] yet? You might end up changing what this constraint does whilst implementing that change, so may as well wait until after that change to decide whether to change the name of this constraint.

ZDevelop94: Yes I have but it wasn't much of a change I just took out one of the case matches in the constraint. If I change it back to nonEmptyTelephoneNumber I will have to come up with different names for the other variables. I think the current name reflects what it does as well as keeping consistency.

PR #54, opened 2017-06-02
[ZF,TA][APB-870] Moved no enrolments check to after confirm agency page

tev1622:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -59,13 +59,8 @@ class CheckAgencyController @Inject()
   }
 
   val showCheckAgencyStatus: Action[AnyContent] = AuthorisedWithSubscribingAgent {
```
arturopala: imho we should use here another builder, not forcing enrolments read from auth when not used

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyControllerISpec.scala
```diff
@@ -56,21 +56,33 @@ class CheckAgencyControllerISpec extends BaseControllerISpec with SessionDataMis
     }
 
     "redirect to already subscribed page if user has already subscribed to MTD" in {
+
```
PaulHodgson: These vals are shared between 2 tests - make them common.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
tev1622: We need everything that is provided in AuthorisedWithSubscribingAgent. If we create another builder it will need everything in the existing method except for the one line below, do you think it is necessary producing duplicate code that excludes the enrolment check shown in the code below? 
  ` case true => enrolments map { e => body(authContext)(AgentRequest(e, request)) }`
`

arturopala: W e have to figure out the best possible impl, at the same time ruling out unnecessary auth call because:
- it makes additional load to auth
- it takes additional time

PR #56, opened 2017-06-07
[TA,ZF][APB-870] Move the enrolments check to when agent is not suscr…

tev1622:
…ibed


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -108,11 +111,7 @@ class CheckAgencyController @Inject()
             registrationName = knownFactsResult.taxpayerName,
```
arturopala: Shouldn't we use `AuthorisedAgentWithEmptyEnrolment` here? we do not use `enrolment` anymore.

ZDevelop94: I don't see how? Have you commented on the right part of the code?

PR #58, opened 2017-06-13
[DK][APB-895] signed-out, fix integration tests respectively.

dimitraka71:


dimitraka71:
SignOutControllerISpec is fixed

PR #60, opened 2017-06-14
[APB-805][RV] Added validation for blacklisted postcodes in subscription details page

rvijayan:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
```diff
@@ -41,6 +41,15 @@ package object controllers {
       }
     }
 
+    private def blacklistedPostcode(postcodes: Seq[String]): Constraint[String] = Constraint[String] { fieldValue: String =>
+      val invalidPostcode = postcodes.exists(_.trim == fieldValue.trim)
```
PaulHodgson: Is trim() enough? Is it better to remove all white space, especially in the middle of the postcode? What about case insensitivity?

Probably best to add tests for these cases.

PR #76, opened 2017-07-05
[DK][APB-901] fix error messages and unparsable characters

dimitraka71:


conf/messages
```diff
@@ -1,5 +1,5 @@
-error.postcode.invalid = Please enter a valid postcode
-error.postcode.blacklisted = This postcode is blocked and cannot be used
+error.postcode.invalid = You have entered an invalid postcode
+error.postcode.blacklisted = You can't use the postcode you've entered
```
frankoid: Did this content change come from Charlie? (the old message seems clearer to me but if the change came from Charlie then it's fine)

app/uk/gov/hmrc/agentsubscriptionfrontend/models/Address.scala
```diff
@@ -29,52 +29,73 @@ case class Address(addressLine1: String,
                    postcode: Option[String],
                    countryCode: String)
 
-object Address {
+object Address{
 
   type PostCodeError = Set[String]
```
frankoid: There are some things about the validation code that strike me as a bit strange. Can we pair and try to improve them?

Some examples:
- error type is called `PostCodeError` but surely `Address.validate` may return non-postcode errors too?
- "right" type is Unit, not Address i.e. I'd expect `Validated[PostcodeError, Address]` instead of `Validated[PostCodeError, String]`

conf/messages
dimitraka71: I changed the error messages according to this document: https://docs.google.com/spreadsheets/d/1gIEJJecD_EjBgDlISAcXbpNfQRCIP-SKM5qbdYo7_fQ/edit#gid=0, (subscription-address tab).

app/uk/gov/hmrc/agentsubscriptionfrontend/models/Address.scala
dimitraka71: yes, you are right. I have to fix this

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -290,7 +290,7 @@ class SubscriptionControllerISpec extends BaseControllerISpec with SessionDataMi
         val result = await(controller.submit("addr1")(authenticatedRequest()))
 
         status(result) shouldBe 200
-        checkHtmlResultWithBodyText(result, "This postcode is blocked and cannot be used")
+        checkHtmlResultWithBodyText(result, htmlEscapedMessage("You can't use the postcode you've entered"))
```
frankoid: `htmlEscapedMessage` is intended to take a key in the messages file e.g. `htmlEscapedMessage("error.postcode.blacklisted")`

app/uk/gov/hmrc/agentsubscriptionfrontend/models/Address.scala
frankoid: can we pair on it?

dimitraka71: sure !

PR #78, opened 2017-07-06
[DK, FD][APB-901] decode error messages, refactor, add more tests

dimitraka71:


app/uk/gov/hmrc/agentsubscriptionfrontend/models/Address.scala
```diff
@@ -137,7 +136,4 @@ object Address{
 
     OFormat[Address](reads, formatAddressValue)
   }
-
-  val renderErrors: PostCodeError => String = (errors: PostCodeError) => errors.foldLeft("")(_+", "+_).substring(1)
```
frankoid: Does this have to be a `val`, can it be a `def` instead? `def` is less advanced -> easier to understand

frankoid: Is there a unit test for this? If not please add one

frankoid:
`sbt clean test it:test` fails for me, does it pass for you?

app/uk/gov/hmrc/agentsubscriptionfrontend/models/Address.scala
dimitraka71: according to the renderErrors anonymous function: this is the updated, https://github.com/hmrc/agent-subscription-frontend/pull/78/commits/197c51130aac51d8e9dbef15f2fe731fa42f4be6, (commit 197c51130aac51d8e9dbef15f2fe731fa42f4be6). and all the tests are passed

frankoid: I think my comments still apply - where is the unit test for `renderErrors` and does it need to be a `val`?

dimitraka71: I can write unit tests for this anon function. It has the same functionality with def

PR #79, opened 2017-07-07
[DK][Apb 901]

dimitraka71:


test/uk/gov/hmrc/agentsubscriptionfrontend/models/MessagesValidationSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+import play.api.data.validation.ValidationError
+
+class MessagesValidationSpec extends PlaySpec with OneAppPerSuite {
+
+  "renderErrors function" should {
+    "concatenate invalid and blacklist error messages" in {
+      val validationErrors = Set(ValidationError("error.postcode.invalid"), ValidationError("error.postcode.blacklisted"))
+      val result: String = renderErrors(validationErrors)
+
+      result mustEqual " You have entered an invalid postcode, You can't use the postcode you've entered"
```
frankoid: Ah, now I understand what the `substring(1)` in `renderErrors` is there for :)

It would be nice to avoid the leading space, I will push a commit that does this.

frankoid: It's best to use should matchers (`shouldEqual` instead of `mustEqual`) for consistency with our other tests.

dimitraka71:
@frankoid, I use fold in order to merge the error messages separated with comma. In order to ignore the first comma, I use substring.

@frankoid, Matchers from FlatSpec have conflicting methods with PlaySpec.

frankoid:
> @frankoid, I use fold in order to merge the error messages separated with comma. In order to ignore the first comma, I use substring.

Yes, I figured that out. I think it's cleaner to use `reduceOption` and `getOrElse` than `foldLeft` and `substring`. Also I don't see any reason to include the leading space.

> @frankoid, Matchers from FlatSpec have conflicting methods with PlaySpec.

I know, but I don't see a strong reason to use `PlaySpec`, all it is is a convenience superclass that mixes a few traits. We can easily pick and choose from those traits and mix the ones we want into individual tests, or make our own convenience superclasses with selections of traits (see for example ). If we were starting from scratch I can see why we'd use it, but since we have lots of existing code that uses should matchers I think it's better to follow that convention.

frankoid:
P.S. I have already made the suggested changes so no need for you to make them too!

dimitraka71:
reduceOption with getOrElse is better solution

PR #80, opened 2017-07-10
Assets rebased

somme:


frankoid:
I thought that assets usually went into assets-frontend on MDTP (not into individual frontend services)?

frankoid:
Is there a JIRA that explains what this work is about?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -290,7 +290,7 @@ class SubscriptionControllerISpec extends BaseControllerISpec with SessionDataMi
         val result = await(controller.submit("addr1")(authenticatedRequest()))
 
         status(result) shouldBe 200
-        checkHtmlResultWithBodyText(result, htmlEscapedMessage("error.postcode.blacklisted"))
+        bodyOf(result) should include("Sorry, the address is not in a format that our system can accept. Please change the address to fix the problems described below:")
```
frankoid: Using `checkHtmlResultWithBodyText` has the advantage that the tests won't need to be updated if the message is changed.

frankoid:
Please can we discuss this before it's merged

app/uk/gov/hmrc/agentsubscriptionfrontend/views/helpers/input.scala.html
```diff
@@ -0,0 +1,100 @@
+@*
```
frankoid: This seems to be a fork of input.scala.html from play-ui. It would be nice to use the play-ui version instead.
If that's not possible we should make sure we don't end up with a copy of it in each of our frontends because that would be hard to maintain - we can put it in a different library.

project/Gulp.scala
```diff
@@ -0,0 +1,61 @@
+import play.sbt.PlayRunHook
```
frankoid: What does this do? How should it be used - does it kick in automatically when you run `sbt test` etc or do you have to run `sbt gulp` explicitly. Does anything need to be installed?

Worth updating the README

We should also decide whether we want to serve assets from a dedicated frontend.

I've just tried `sbt test` and it seems this does kick in automatically - and fails on my machine even though I have node etc installed (wrong version?). This might be a reason to put it in its own repository (so that only devs working on the frontend assets need to have JavaScript tools installed).

frankoid:
Closing for now as it breaks `sbt test` on my machine and we need to figure out the design for where the assets will live if we want to go for newer GDS assets (as previously discussed).

PR #83, opened 2017-07-20
[APB-928][AO,JR,RV] Add dependancy injection, tests to the KnownFactsResultMongoRepository

rvijayan:


rvijayan:
Since this is a spike story, no need to merge it with master.

PR #85, opened 2017-07-25
[APB-1004][AO,RV] Fix wrong UTR subscription of clients in agent-subscription-frontend

rvijayan:


app/uk/gov/hmrc/agentsubscriptionfrontend/service/SessionStoreService.scala
```diff
@@ -33,6 +33,12 @@ class SessionStoreService @Inject() (sessionCache: SessionCache) {
   def cacheKnownFactsResult(knownFactsResult: KnownFactsResult)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Unit] =
     sessionCache.cache("knownFactsResult", knownFactsResult).map(_ => ())
 
+  def fetchInitialDetails(implicit hc: HeaderCarrier): Future[Option[InitialDetails]] =
```
frankoid: Please can you add tests for cache/fetchInitialDetails to `SessionStoreServiceSpec`. Hopefully this will be easy to do as the tests can be similar to the existing tests for cache/fetchInitialDetails

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -342,6 +342,19 @@ class SubscriptionControllerISpec extends BaseControllerISpec with SessionDataMi
...
+        AuthStub.hasNoEnrolments(subscribingAgent)
+        sessionStoreService.knownFactsResult = Some(myAgencyKnownFactsResult)
+        sessionStoreService.initialDetails = None
+
+        givenAddressLookupReturnsAddress("addr1")
+        val result = await(controller.submit("addr1")(authenticatedRequest()))
+
+        status(result) shouldBe 303
+        redirectLocation(result).head shouldBe routes.SubscriptionController.showSubscriptionDetails().url
+        sessionStoreService.removeCalled shouldBe true
```
frankoid: If `remove()` has been called then `showSubscriptionDetails` will fail when it tries to `fetchKnownFactsResult` and will just end up calling `sessionMissingRedirect()`. Why not just short circuit this and call `sessionMissingRedirect()` in the first place.

I'm also not sure this test is in the right place (inside `"redisplay form" when`) - all the other tests inside `"redisplay form" when` are redisplaying the form, including data, after a validation error but this isn't what happens in this case.

See `SessionControllerISpec."redirect to the Check Agency Status page if there is no KnownFactsResult in session because the user has returned to a bookmark"` for an example of testing session missing functionality.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -113,28 +106,40 @@ class SubscriptionController @Inject()
...
                 }
+                _ <- sessionStoreService.remove()
+              } yield subscriptionResponse
+
+              subscriptionResponse.map {
+                case Right((arn, agencyName)) => Redirect(routes.SubscriptionController.showSubscriptionComplete())
+                  .flashing("arn" -> arn.arn, "agencyName" -> agencyName)
+                case Left(CONFLICT) => Redirect(routes.CheckAgencyController.showAlreadySubscribed())
+                case Left(FORBIDDEN) => Redirect(routes.SubscriptionController.showSubscriptionFailed())
+                case Left(MISSING_SESSION_DATA) => Redirect(routes.SubscriptionController.showSubscriptionDetails())
```
frankoid: I'm not keen on the same val being able to contain two different things like this (either an HTTP status code returned by agent-subscription or a magic number to represent missing session data). It's unclear and error-prone.

How about using a sealed trait as the content of the Left. So `subscriptionResponse` would be a `Future[Either[SubscriptionFailed, (Arn, String)]]`.
```
  private sealed trait SubscriptionFailed
  private case class SubscriptionReturnedHttpError(httpStatusCode: Int) extends SubscriptionFailed
  private case object MissingSessionData extends SubscriptionFailed
```

frankoid:
I think that it is "too magic" for `SubscriptionDetails.mapper` to be implicit - and also unnecessary at least after the changes in this PR (code compiles & tests pass for me without it). Shall we remove the `implicit` keyword from `implicit def mapper(`... ?

PR #88, opened 2017-07-28
[APB-991][JR,RV,AO] redirect to the new sos journey for agents and preserve the known facts in the mongodb

rvijayan:


arturopala:
This is the result of the APB-928 spike followed by the APB-991 refine/integrate story

app/uk/gov/hmrc/agentsubscriptionfrontend/GuiceModule.scala
```diff
@@ -35,6 +36,45 @@ class GuiceModule(environment: Environment, configuration: Configuration) extend
...
+  trait ConfigPropertyType[A] {
+    def bindConfigProperty(clazz: Class[A])(propertyName: String): ScopedBindingBuilder
+  }
+
+  object ConfigPropertyType {
+    implicit val stringConfigProperty = new ConfigPropertyType[String] {
+      def bindConfigProperty(clazz: Class[String])(propertyName: String): ScopedBindingBuilder =
+        bind(clazz).annotatedWith(Names.named(s"$propertyName")).toProvider(new StringConfigPropertyProvider(propertyName))
+
+      private class StringConfigPropertyProvider(propertyName: String) extends Provider[String] {
```
johnrevill: This class is very similar to the `IntConfigPropertyProvider` below on line 61.
It seems at first glance the only difference between the two being is the `Configuration.getString/getInt` call and whether the class deals with ints or strings. 
Could a generic class be used instead? e.g. parameterized with either int/string types and provided with a Configuration function to call. 
I'm guessing it might end up looking too complicated to gain any real benefit though..

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignedOutController.scala
```diff
@@ -16,22 +16,41 @@
...
+      URLEncoder.encode(s"${routes.StartController.returnAfterGGCredsCreated().absoluteURL()}$query", "UTF-8")
+
+    for {
+      knownFactOpt <- sessionStoreService.fetchKnownFactsResult
+      id <- knownFactOpt match {
+        case Some(x) => knownFactsResultMongoRepository.create(x).map(Option.apply)
+        case None => Future.successful(None)
+      }
+    } yield {
+      val continueUrl = returnAfterGGCredsCreatedUrl(id.map(i => s"?id=$i").getOrElse(""))
```
johnrevill: If we failed to persist to the DB, should we instead be showing an error page? 
Currently in this case it looks like the user would be redirected to GG on the SOS journey but they'll return without a valid ID.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignedOutController.scala
```diff
@@ -16,22 +16,41 @@
...
+      Redirect(s"$sosUrl&continue=$continueUrl", 303).withNewSession
+    }
   }
 
-  def startSurvey = Action.async { implicit request =>
-    successful(Redirect(surveyUrl, 303).withNewSession)
+  def startSurvey = Action { implicit request =>
+    Redirect(surveyUrl, 303).withNewSession
```
johnrevill: Potential for using the descriptive 'Results.SeeOther(surveyUrl)' instead of Redirect(surveyUrl, 303). Both here and on line 49 above

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignedOutController.scala
arturopala: We are also redirecting to SOS in non-agent scenario where known facts has not been asked

app/uk/gov/hmrc/agentsubscriptionfrontend/GuiceModule.scala
arturopala: The limitation comes from Typesafe's Config library not supporting truly generic getters. IMHO current state of property providers is a balance between verbosity vs readability.

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignOutControllerISpec.scala
```diff
@@ -1,22 +1,54 @@
...
+import uk.gov.hmrc.agentmtdidentifiers.model.Utr
+import uk.gov.hmrc.agentsubscriptionfrontend.models.KnownFactsResult
+import uk.gov.hmrc.agentsubscriptionfrontend.repository.KnownFactsResultMongoRepository
+
+import scala.concurrent.ExecutionContext.Implicits.global
 
 class SignOutControllerISpec extends BaseControllerISpec {
   private lazy val controller: SignedOutController = app.injector.instanceOf[SignedOutController]
 
   private lazy val logoutRedirectUrl = "/government-gateway-registration-frontend/choose-your-account"
```
johnrevill: This `logoutRedirectUrl` val now looks unused as it has been replaced by `sosRedirectUrl`

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
```diff
@@ -1,14 +1,21 @@
 package uk.gov.hmrc.agentsubscriptionfrontend.controllers
 
 import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.mvc.Result
```
johnrevill: (pedantic) Unused import

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
```diff
@@ -56,7 +63,7 @@ class StartControllerISpec extends BaseControllerISpec {
       val result = await(controller.showNonAgentNextSteps(authenticatedRequest()))
 
       status(result) shouldBe 200
-      bodyOf(result) should include("/signed-out")
+      bodyOf(result) should include("/redirect-to-sos")
```
johnrevill: The test this is part of is "allow the government gateway URL to be configured". Not sure if the name of this test still makes sense.

PR #93, opened 2017-08-03
[CC][APB-1019] Fix scoverage excluded packages

cavster:
Fixed regex for coverageExcludedPackages so it will ignore routes filesetc...


frankoid:
doesn't actually upgrade scoverage plugin

PR #97, opened 2017-08-09
[DK][refactor-validated-type] replace nonEmptyList with validatedNel,…

dimitraka71:
… rewrite optionInside method

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/models/AddressValidator.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/models/AddressValidator.scala
@@ -17,7 +17,7 @@
 package uk.gov.hmrc.agentsubscriptionfrontend.models
 
 import cats.data.Validated.{Invalid, Valid}
-import cats.data.{NonEmptyList, Validated}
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/models/AddressValidator.scala
```diff
@@ -28,74 +28,83 @@ import scala.util.matching.Regex
 
 object AddressValidator {
 
-  type ValidationErrors = NonEmptyList[ValidationError]
+  type ValidatedType[A] = ValidatedNel[ValidationError, A]
```
frankoid: I think that `ValidatedType` makes things less clear - I think it should be inlined

dimitraka71: We can either define validatedType for each type,
val validatedDesAddress = ValidatedNel[ValidationError, DesAddress]
val validatedLine = ValidatedNel[ValidationError, String]
etc...
or this generic validatedType[A]

PR #98, opened 2017-08-09
APB-1055 Remove agent-subscription/not-subscribed

arturopala:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -35,6 +35,15 @@ import uk.gov.hmrc.play.frontend.controller.FrontendController
 
 import scala.concurrent.Future
 
+object CheckAgencyController {
```
ZDevelop94: not sure about this object, greatly confuses things and it's use is only for the knownfacts form. I recommend putting back into main controller and using `val controller = new CheckAgencyController(Dependency,Dependency)`to test the knowfacts form. Or at least change the object name?

arturopala: I think this is established pattern in Scala to keep public final immutable stuff in companion object having same name as a class? KnownFactsForm doesn't depend on anything else in controller so we should keep testing it simple as possible.

ZDevelop94: oh okay that sounds understandable.

PR #100, opened 2017-08-11
[DK, LS][APB-1041] modify configuration for AS account to meet the ne…

dimitraka71:
…w standard, update scalatestplus version.

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -164,7 +164,7 @@ Prod {
       }
 
       agent-services-account-frontend {
-        external-url = "localhost:9401"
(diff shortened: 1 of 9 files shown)


conf/application.conf
```diff
@@ -153,6 +163,11 @@ Prod {
         port = 80
       }
```
PaulHodgson: This will never be 'localhost' for any non-local environment. Perhaps safest to put the prod url as a default? At least that way it works without further configuration in production. The same point applies to the address-lookup config too.

dimitraka71: fixed

PR #101, opened 2017-08-11
[CC][APB-1014]Refactored  AddressValidator to a class so we can test …

cavster:
…the logging minor refactoring to SubscriptionController and test helpers added

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/GuiceModule.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/GuiceModule.scala
@@ -23,7 +23,7 @@ import com.google.inject.AbstractModule
 import com.google.inject.binder.ScopedBindingBuilder
 import com.google.inject.name.Names
 import play.api.Mode.Mode
-import play.api.{Configuration, Environment}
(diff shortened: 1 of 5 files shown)


frankoid:
LGTM @cavster please merge if you are happy with the changes I made in https://github.com/hmrc/agent-subscription-frontend/pull/101/commits/a660e75ac6d1702e55d339575c1ab392e66c81d7

frankoid:
I think there is potential to improve the code further (but in another PR, I think it's best to merge this one first as is) by making `SubscriptionService.subscribeAgencyToMtd` return errors in a less leaky-abstraction way by returning  a `Future[Either[SubscriptionError, Arn]]` instead of a `Future[Either[Int, Arn]]`.

`SubscriptionError` would be a sealed trait with `SubscriptionForbidden` and `SubscriptionConflict` concrete subtypes - and then `SubscriptionError` could also be used in `SubscriptionController` instead of `SubscriptionReturnedHttpError`

OR we could go the other way altogether and make `SubscriptionService.subscribeAgencyToMtd` return a `Future[Either[Upstream4xxResponse, Arn]]` and then use `Upstream4xxResponse` instead of `SubscriptionReturnedHttpError` in the controller.

What do you think?

PR #106, opened 2017-08-16
[Dk, LS][auth-bug-fix-1045] refresh enrolments

dimitraka71:


conf/application.conf
```diff
@@ -119,6 +124,11 @@ Test {
         start.path = "agent-services-account"
       }
 
+      government-gateway-authentication {
```
cavster: why is it in here 3 times?

dimitraka71: dev, test, prod environments

frankoid:
Where are the tests for this code?

dimitraka71:
@frankoid , you are right. tests are missing at the moment. We merged it to master this morning in order to make sure that refreshment of enrolments works. We will add the tests.

conf/application.conf
cavster: Shouldn't it be different  for the different environment's  dev should be www-dev.tax.service.gov.uk etc.. And why not put in the appropriate app-configs ? and if they are the same you can put in app-config-base ;)

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -187,12 +189,16 @@ class SubscriptionController @Inject()
...
       implicit request => {
-        Future successful{
+        Future successful {
           val agencyData = for {
             agencyName <- request.flash.get("agencyName")
             arn <- request.flash.get("arn")
           } yield (agencyName, arn)
 
+          authenticatorConnector.refreshEnrolments.map { httpResponse =>
+            Logger.info(">>>>> AUTHENTICATOR RESPONSE: " + httpResponse.status)
```
cavster: Be nice to have a unit test ;)

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AuthenticatorConnector.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import play.api.Logger
+import uk.gov.hmrc.play.http.{HeaderCarrier, HttpGet, HttpPost, HttpResponse}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuthenticatorConnector @Inject()(@Named("government-gateway-authentication-baseUrl") baseUrl: URL, http: HttpGet with HttpPost) {
+
+  def refreshEnrolments(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] = {
+    http.POSTEmpty(s"$baseUrl/government-gateway-authentication/refresh-profile") map { response =>
```
cavster: I am guessing POSTEmpty is a new thing with http-verbs?  Also we are meant to be using MdcLoggingExecutionContext  instead of ExecutionContext ;)

cavster: Also it would nice to have a small IT or unit tests.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
cavster: Hmmmm better to use for each if it's a side effect ;)

cavster: You could also put everything here in the same for yield statement ;) just use _ = for the logging thing

conf/application.conf
dimitraka71: dev is for local environment, this is how is defined from play framework. The Prod environment is the actual profile configuration that is being overwritten by app-config-*

PR #118, opened 2017-08-22
[APB-1002][AO] fix issue with missing continue param when redirecting to AS page

arturopala:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -200,6 +199,7 @@ class SubscriptionController @Inject()
             case Some((agencyName, arn)) =>
               sessionStoreService.fetchContinueUrl.
                 recover { case NonFatal(_) => None }.
```
frankoid: Exception not logged 😱

frankoid: If we were catching a specific exception here that we considered "normal" then it'd be OK not to log it, but for a case as broad as `NonFatal` we should be logging it.

I know this code wasn't added in this PR but just saw it and I think it's important for us to have good production logging.

arturopala: Fixed.

PR #121, opened 2017-08-29
[TA][APB-1112] Added DesAddressForm validation to the SubscriptionCon…

tev1622:
…troller

* Removed unnecessary methods and pages that are no longer in use
* Stopped redirecting to the address lookup service after acknowledgement of errors


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -219,9 +233,15 @@ class SubscriptionController @Inject()
...
 
-  def renderErrors(errors: NonEmptyList[ValidationError])(implicit messages: Messages): String = errors
-    .toList
-    .map(valError => Messages(valError.message, valError.args: _*))
-    .reduce(_ + ", " + _)
-
+  def renderErrors(errors: NonEmptyList[ValidationError])(implicit messages: Messages): String =
+    errors
+      .toList
+      .map(valError => Messages(valError.message, valError.args: _*))
```
cavster: Dido

frankoid:
Coverage has gone down from:

Statement coverage: 96.26% | Branch coverage: 94.74% in master
to
Statement coverage: 95.71% | Branch coverage: 90.91% in this branch

I think we should get coverage back up to the previous level, or at least understand why it has fallen and be OK with the reason.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -130,56 +130,70 @@ class SubscriptionController @Inject()
...
-                    res ← subscribe(details, desAddress, address)
-                  } yield res
-                  subscriptResponse.map(redirectSubscriptionResponse)
+              val addressForm = desAddressForm.bindAddressLookupFrontendAddress(details.utr, address)
+              if (addressForm.hasErrors) {
+                Future successful
+                  Ok(html.address_form_with_errors(addressForm, SubscriptionController.renderDesAddressErrors(addressForm.errors)))
+              } else {
+                val subscriptResponse = for {
+                  res ← subscribe(details, addressForm.get, Some(address))
```
cavster: I am not a fan of using a for yield for one statement but ¯\_(ツ)_/¯

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -130,56 +130,70 @@ class SubscriptionController @Inject()
...
 
-        def redirectSubscriptionResponse(either: Either[SubscriptionReturnedHttpError, (Arn, String)]): Result = {
-          either match {
-            case Right((arn, agencyName))                      => Redirect(routes.SubscriptionController.showSubscriptionComplete())
-              .flashing("arn" -> arn.arn, "agencyName" -> agencyName)
-            case Left(SubscriptionReturnedHttpError(CONFLICT)) => Redirect(routes.CheckAgencyController.showAlreadySubscribed())
-            case Left(SubscriptionReturnedHttpError(_))        => Redirect(routes.SubscriptionController.showSubscriptionFailed())
-          }
-        }
+  def returnFromAddressLookup(id: String): Action[AnyContent] = AuthorisedWithSubscribingAgentAsync() {
```
cavster: Also the name returnFromAddressLookup it's not really what it's doing It is returning the address turning into a form validating it and redirecting based on the data so yeh

app/uk/gov/hmrc/agentsubscriptionfrontend/views/address_form_with_errors.scala.html
```diff
@@ -0,0 +1,73 @@
+@import uk.gov.hmrc.agentsubscriptionfrontend.config.AppConfig
+@import uk.gov.hmrc.play.views.html.helpers.form
+@import uk.gov.hmrc.play.views.html.helpers.input
+@import uk.gov.hmrc.agentsubscriptionfrontend.views.html.formErrorSummary
+@import uk.gov.hmrc.agentsubscriptionfrontend.controllers.routes.SubscriptionController
+@import uk.gov.hmrc.agentsubscriptionfrontend.controllers.routes
+@import uk.gov.hmrc.agentsubscriptionfrontend.models.DesAddress
+@(detailsForm: Form[DesAddress], message: String)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
```
cavster: also rename name message to error messages to be a little clearer to what they are referring too ;)

app/uk/gov/hmrc/agentsubscriptionfrontend/views/address_form_with_errors.scala.html
```diff
@@ -0,0 +1,73 @@
...
+
+@uk.gov.hmrc.agentsubscriptionfrontend.views.html.main_template(appConfig, title = Messages("subscriptionDetails.title"), bodyClasses = None) {
+
+<header class="page-header">
+    <h1>@Messages("invalidAddress.title")</h1>
+</header>
+
+<p>@Messages("invalidAddress.p1")</p>
+
+<p>@message</p>
```
cavster: I don't think this is needed

app/uk/gov/hmrc/agentsubscriptionfrontend/views/address_form_with_errors.scala.html
cavster: Also it turns out we don't need this but double check with charlie!

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
frankoid: The name `returnFromAddressLookup` is based on this being used when returning from address-lookup-frontend i.e. this is where address-lookup-frontend redirects you to after you've picked an address.

It does do a few different things so is quite difficult to name, I can't think of anything better than `returnFromAddressLookup`.

Any suggestions?

cavster: It's fine the way it is  ¯_(ツ)_/¯

frankoid:
LGTM - just needs someone to review the commit I did without a pair and can then be merged @tev1622 @cavster

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -130,56 +129,67 @@ class SubscriptionController @Inject()
...
+        desAddressForm.form.bindFromRequest().fold(
+          formWithErrors => {
+            Future successful Ok(html.address_form_with_errors(formWithErrors))
+          },
+          validDesAddress =>
+            sessionStoreService.fetchInitialDetails.flatMap { maybeInitialDetails =>
+              maybeInitialDetails.map { initialDetails =>
+                val subscriptResponse = for {
+                  res ← subscribe(initialDetails, validDesAddress, None)
+                  _ ← sessionStoreService.remove()
```
frankoid: Looks like this call to `sessionStoreService.remove()` might break the continueUrl functionality - it relies on the continue URL being available in `showSubscriptionComplete` which happens after this.

frankoid:
actually see https://github.com/hmrc/agent-subscription-frontend/pull/121/files#r136319602 - that needs checking out before we merge. @arturopala @johnrevill is there a test (e.g. end-to-end test) that I can run to see if this breaks the APB-1002 continue URL functionality or will I need to test it manually?

frankoid:
@tev1622 @cavster I have removed that call to `sessionStoreService.remove()` because it did break the continue URL functionality, so this should be good to go now.

PR #124, opened 2017-09-06
[FD][APB-1014,APB-1288] Quick fix for behaviour change introduced in APB-1014

frankoid:
Prior to APB-1014 if the user chose a non-UK country we ignored their choice and sent "GB" to the DES subscription API anyway.
APB-1014 changed that so now we send the country sent by the user.
This commit changes the behaviour back so we always send "GB".

This is an interim fix - when we do APB-1288 we will inform the user that non-UK addresses are not allowed instead of silently ignoring their choice of country.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
@@ -18,6 +18,7 @@ package uk.gov.hmrc.agentsubscriptionfrontend.service
 
 import javax.inject.{Inject, Singleton}
 
+import play.api.Logger
 import play.api.http.Status
 import uk.gov.hmrc.agentsubscriptionfrontend.connectors.AgentSubscriptionConnector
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -31,11 +32,19 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
 
   def subscribeAgencyToMtd(subscriptionDetails: SubscriptionDetails)
                           (implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[Int, Arn]] = {
+    val address = if (subscriptionDetails.address.countryCode != "GB") {
+      Logger.warn(s"Non-GB country code chosen by user for UTR ${subscriptionDetails.utr.value}. " +
+                  s"Overriding with GB. A better fix for this is coming in APB-1288.")
```
cavster: Just say every address is a English address ¯\_(ツ)_/¯

PR #151, opened 2018-01-04
[APB-1779][AK] error messages fix, connector fail case fix

arkowalski:


app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
```diff
@@ -47,9 +47,9 @@ class AgentAssuranceConnector @Inject()(@Named("agent-assurance-baseUrl") baseUr
   def getActiveCesaRelationship(url: String)(implicit hc: HeaderCarrier): Future[Boolean] = {
     monitor(s"ConsumedAPI-AgentAssurance-getActiveCesaRelationship-GET") {
       http.GET[HttpResponse](baseUrl + url).map(
-        response => response.status == 200)
```
rvijayan: may be, we should keep this check, just to make sure we get true only when the api returns 200.

arturopala: it is already checked in `handleResponse` method

PR #168, opened 2018-03-06
[APB-1316][HO,AK] added ability to block blacklisted utrs

habeeb01:


app/uk/gov/hmrc/agentsubscriptionfrontend/GuiceModule.scala
```diff
@@ -113,6 +113,7 @@ class GuiceModule(environment: Environment, configuration: Configuration) extend
     bindConfigProperty[String]("sosRedirectUrl")
     bindConfigProperty[Int]("mongodb.knownfactsresult.ttl")
     bindConfigProperty[Boolean]("agentAssuranceFlag")
+    bindConfigProperty[String]("r2dw-agentsAssurance-key")
```
arturopala: what is the reason to have multiple r2dw lists?

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
```diff
@@ -55,6 +57,19 @@ class AgentAssuranceConnector @Inject()(@Named("agent-assurance-baseUrl") baseUr
     }
   }
 
+  def isR2DWAgent(utr: Utr)(implicit hc: HeaderCarrier): Future[Boolean] = {
```
arturopala: What about having an endpoint to check by UTR and receive 204/404? It would be more efficient then ask for a list each time.

arkowalski: this followed the instructions from the use case, although we had a discussion with Martin and we will be changing this to a yes/no endpoint

app/uk/gov/hmrc/agentsubscriptionfrontend/GuiceModule.scala
arkowalski: There is just one collection, with multiple Properties(key, value), where the key represents different list.
so r2dw-agentsAssurance-key: r2dw list
and there will be a different key for MAA list

arturopala: but that information should be implicit, as I understand we've got 2 set of endpoints, one for r2dw and other for manually assured?

PR #173, opened 2018-03-12
[APB-2013][JR] Allow agents on the manually assured agents list.

johnrevill:
Also has a small refactor to extract out some agent assurance logic from the CheckAgencyController into a new AssuranceService.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
@@ -71,6 +71,20 @@ class AgentAssuranceConnector @Inject()(@Named("agent-assurance-baseUrl") baseUr
     }
   }
 
+  def isManuallyAssuredAgent(utr: Utr)(implicit hc: HeaderCarrier): Future[Boolean] = {
+    monitor(s"ConsumedAPI-AgentAssurance-getManuallyAssuredAgents-GET") {
+      val endpoint = s"/agent-assurance/manually-assured/${utr.value}"
+      http.GET[HttpResponse](new URL(baseUrl, endpoint).toString).map{ response =>
+        (200 until 300) contains response.status
+      }.recover {
+        case e: Upstream4xxResponse if e.upstreamResponseCode == 403 => false
+        case _: NotFoundException => {
+          throw new IllegalStateException(s"unable to reach $baseUrl/$endpoint. Manually assured agents list might not have been configured")
+        }
+      }
+    }
+  }
+
   private def cesaGetUrl(ninoOrUtr: String, valueOfNinoOrUtr: String, saAgentReference: SaAgentReference): String = {
     s"/agent-assurance/activeCesaRelationship/$ninoOrUtr/$valueOfNinoOrUtr/saAgentReference/${saAgentReference.value}"
   }
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
@@ -28,7 +28,8 @@ import uk.gov.hmrc.agentmtdidentifiers.model.Utr
 import uk.gov.hmrc.agentsubscriptionfrontend.audit.AuditService
 import uk.gov.hmrc.agentsubscriptionfrontend.auth.{AgentRequest, AuthActions, NoOpRegimeWithContinueUrl}
 import uk.gov.hmrc.agentsubscriptionfrontend.config.AppConfig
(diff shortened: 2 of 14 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
```diff
@@ -71,6 +71,20 @@ class AgentAssuranceConnector @Inject()(@Named("agent-assurance-baseUrl") baseUr
     }
   }
 
+  def isManuallyAssuredAgent(utr: Utr)(implicit hc: HeaderCarrier): Future[Boolean] = {
+    monitor(s"ConsumedAPI-AgentAssurance-getManuallyAssuredAgents-GET") {
+      val endpoint = s"/agent-assurance/manually-assured/${utr.value}"
+      http.GET[HttpResponse](new URL(baseUrl, endpoint).toString).map{ response =>
+        (200 until 300) contains response.status
```
arturopala: I feel it might not be an efficient method for checking

app/uk/gov/hmrc/agentsubscriptionfrontend/service/AssuranceService.scala
```diff
@@ -0,0 +1,92 @@
...
+        if(isManuallyAssured)
+          Future.successful(None)
+        else {
+          val futurePaye = agentAssuranceConnector.hasAcceptableNumberOfPayeClients
+          val futureSA = agentAssuranceConnector.hasAcceptableNumberOfSAClients
+
+          for {
+            hasAcceptableNumberOfPayeClients <- futurePaye
+            hasAcceptableNumberOfSAClients <- futureSA
+          } yield Some(AssuranceResults(hasAcceptableNumberOfPayeClients, hasAcceptableNumberOfSAClients))
```
arturopala: It might be better to have 'AssuranceResults' containing all decisions made, including manually assured and r2dw

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -154,7 +137,11 @@ class CheckAgencyController @Inject()
         }
       }
     }
-    if(agentAssuranceFlag) processCheckAgencyStatusWithR2DW else processCheckAgencyStatus
+
+    assuranceService.isOnRefusalToDealWithList(knownFacts.utr) flatMap {
```
arturopala: could we make it part of `assuranceService.assureIsAgent`?

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
johnrevill: Just double-checked this at http://docs.scala-lang.org/overviews/collections/concrete-immutable-collection-classes.html#ranges : "Ranges are represented in constant space, because they can be defined by just three numbers: their start, their end, and the stepping value. Because of this representation, most operations on ranges are extremely fast."

The contains() function just checks the bounds of the range: https://github.com/scala/scala/blob/07d61ecf134dbba143531f5a1bd3c1289c437296/src/library/scala/collection/immutable/Range.scala#L336

app/uk/gov/hmrc/agentsubscriptionfrontend/service/AssuranceService.scala
johnrevill: Sounds good, I'll change it.

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentAssuranceConnector.scala
arturopala: that's fine

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
johnrevill: Yep, and mentioned in the AssuranceResults

app/uk/gov/hmrc/agentsubscriptionfrontend/models/AssuranceResults.scala
```diff
@@ -16,6 +16,34 @@
 
 package uk.gov.hmrc.agentsubscriptionfrontend.models
 
-case class AssuranceResults(
-                             hasAcceptableNumberOfPayeClients: Boolean,
-                             hasAcceptableNumberOfSAClients: Boolean)
+case class AssuranceResults(isOnRefusalToDealWithList: Boolean,
+                            isManuallyAssured: Boolean,
```
arturopala: Should `isManuallyAssured` be also an `Option[Boolean]`?

PR #206, opened 2018-07-26
[APB-2816][AK] cater for partially subscribed agents, when HMRC-AS-AGENT enrolment is not allocated

arkowalski:


app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -20,17 +20,38 @@ import javax.inject.{Inject, Singleton}
...
 import uk.gov.hmrc.agentsubscriptionfrontend.models._
 import uk.gov.hmrc.http.{HeaderCarrier, Upstream4xxResponse}
 
 import scala.concurrent.{ExecutionContext, Future}
 
+case class SubscriptionReturnedHttpError(httpStatusCode: Int) extends Product with Serializable
+
+object SubscriptionState extends Enumeration {
+  type SubscriptionState = Value
+  val BrandNewSubscription, IsOnlySubscribedInETMP, IsSubscribedToAgentServices, NoRegistrationFound = Value
```
rvijayan: What does "NoRegistrationFound" mean? Is it " no subscription found in etmp"?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
 
-    agentSubscriptionConnector.getRegistration(knownFacts.utr, knownFacts.postcode) flatMap {
-      case Some(Registration(Some(taxpayerName), isSubscribedToAgentServices)) if !isSubscribedToAgentServices =>
-        processCheckAgencyStatus(knownFacts.utr, taxpayerName, isSubscribedToAgentServices)
-      case Some(Registration(_, isSubscribedToAgentServices)) if isSubscribedToAgentServices =>
+    subscriptionService.getSubscriptionStatus(knownFacts.utr, knownFacts.postcode).flatMap {
+      case SubscriptionProcess(SubscriptionState.BrandNewSubscription, Some(registrationDetails)) =>
+        processCheckAgencyStatus(
+          knownFacts.utr,
+          registrationDetails.taxpayerName.get,
```
rvijayan: Will `taxpayerName.get` throw an exception if its None? Cant see where we are checking non-emptiness.

arkowalski: you need to be: SubscriptionState.BrandNewSubscription
otherwise you be protected by this:
case Some(Registration(None, _, _)) =>
        throw new IllegalStateException(s"The agency with UTR ${utr.value} has a missing organisation/individual name.")

arkowalski: I can put getOrElse there as well to make it clear if you want

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
+          registrationDetails.isSubscribedToAgentServices)
+      case SubscriptionProcess(SubscriptionState.IsOnlySubscribedInETMP, Some(registrationDetails)) =>
+        agent match {
+          case hasNonEmptyEnrolments(_) =>
+            Future successful Redirect(routes.CheckAgencyController.showHasOtherEnrolments())
```
rvijayan: Do we need this redirect here? I cant see this requirement in the ticket, neither this has been on the previous version..

arkowalski: https://jira.tools.tax.service.gov.uk/secure/attachment/138239/Subscription%20Journey%20-%20APB-2582.pdf

this is the graph attached with the ticket, that was updated by John

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
+          case _ => {
+            subscriptionService
+              .completePartialSubscription(
+                CompletePartialSubscriptionBody(knownFacts.utr, SubscriptionRequestKnownFacts(knownFacts.postcode)))
```
johnrevill: Not sure if the controller should need to know or care about the CompletePartialSubscriptionBody or SubscriptionRequestKnownFacts classes. 
They're mainly for making calls to the backend with. i.e. the subscriptionService seems to be leaking implementation details in this interface.
Could the service's completePartialSubscription function instead just take a plain Utr (typed) and postcode?

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
arkowalski: yes, getSubscriptionStatus = NoRegistrationFound     in ETMP,
I can rename

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
+                case _ => throw new InternalServerException("partialSubscription fix executed, but failed to complete")
+              }
+          }
+        }
+      case SubscriptionProcess(SubscriptionState.IsSubscribedToAgentServices, _) => {
         mark("Count-Subscription-AlreadySubscribed-RegisteredInETMP")
         Future successful Redirect(routes.CheckAgencyController.showAlreadySubscribed())
-      case Some(_) =>
-        throw new IllegalStateException(
-          s"The agency with UTR ${knownFacts.utr} has a missing organisation/individual name.")
```
rvijayan: Why is this error not relevant now?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
-      case Some(Registration(_, isSubscribedToAgentServices)) if isSubscribedToAgentServices =>
+    subscriptionService.getSubscriptionStatus(knownFacts.utr, knownFacts.postcode).flatMap {
+      case SubscriptionProcess(SubscriptionState.BrandNewSubscription, Some(registrationDetails)) =>
+        processCheckAgencyStatus(
+          knownFacts.utr,
+          registrationDetails.taxpayerName.get,
+          registrationDetails.isSubscribedToAgentServices)
+      case SubscriptionProcess(SubscriptionState.IsOnlySubscribedInETMP, Some(registrationDetails)) =>
+        agent match {
+          case hasNonEmptyEnrolments(_) =>
```
johnrevill: Not suggesting this needs to  be changed, but just a thought for the future.. 
We could use a similar pattern to the  `withSubscribingAgent` function here, e.g.:
```withCleanCreds {
  // Complete partial subscription
}```

..and then within the implementation of `withCleanCreds`, it'd do a redirect to `/has-other-enrolments` if they've got unclean creds.

File: app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala (hunk 4 above)
Comment (arkowalski) [reply app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala]: sure, ill change

File: app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala (hunk 5 above)
Comment (arkowalski) [reply app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala]: its in subscriptionService.getSubscriptionStatus

File: app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala [hunk 7]
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
       case e => throw e
     }
   }
+
+  def completePartialSubscription(details: CompletePartialSubscriptionBody)(
+    implicit hc: HeaderCarrier,
+    ec: ExecutionContext): Future[(Boolean, Option[Arn])] =
+    for {
+      subscriptionProgress <- getSubscriptionStatus(details.utr, details.knownFacts.postcode) //FixMe APB-2805 this check should happen previously and can be stored in the session
```
rvijayan: I think we shouldn't mix getSubscriptionStatus and completePartialSubscription. Identifying if the process is opted for partial subscription should be an independent in if you look at checkAgencyStatusGivenValidForm method in CheckAgencyController, mixing these two methods result in calling getSubscriptionStatus twice for no reason.

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnectorISpec.scala
```diff
@@ -116,6 +130,54 @@ class AgentSubscriptionConnectorISpec extends BaseISpec with MetricTestSupport {
     }
   }
 
+  "completePartialSubscription" should {
```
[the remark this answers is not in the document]
arkowalski: added

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
arkowalski: https://jira.tools.tax.service.gov.uk/browse/APB-2805
will resolve this and the extra call in returnAfterGGCredsCreated

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
+            Future successful Redirect(routes.CheckAgencyController.showHasOtherEnrolments())
+          case _ => {
+            subscriptionService
+              .completePartialSubscription(
```
johnrevill: The `completePartialSubscription` function returns a Boolean indicating whether they were partially subscribed or not.
Does this make the check above for partially subscriptions (via `getSubscriptionStatus`) now redundant, and we could just call `completePartialSubscription` without `completePartialSubscription`?

arkowalski: this process wil be simplified when APB-2805 is implemented as the check will happen once, and we need to check the status to know where to redirect them.

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
...
+      isPartiallySubscribed = subscriptionProgress.state == SubscriptionState.IsOnlySubscribedInETMP
+      (wasPartiallySubscribed, optSucceededAllocatingArn) <- isPartiallySubscribed match {
+                                                              case true =>
+                                                                agentSubscriptionConnector
+                                                                  .completePartialSubscription(details)
+                                                                  .map(arn => (true, Some.apply(arn)))
+                                                                  .recover {
+                                                                    case e: Upstream4xxResponse
+                                                                        if Seq(Status.FORBIDDEN, Status.CONFLICT) contains e.upstreamResponseCode =>
+                                                                      (true, None)
```
rvijayan: Here Status.FORBIDDEN eventually leads us to showAlreadySubscribed page. Is that the correct behaviour? The backend returns FORBIDDEN when there is not ETMP record/ postcodes don't match/ isAnASAgent flag on etmp record is false. I think we should also log these exceptions.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
```diff
@@ -184,18 +186,35 @@ class CheckAgencyController @Inject()(
...
+                CompletePartialSubscriptionBody(knownFacts.utr, SubscriptionRequestKnownFacts(knownFacts.postcode)))
+              .map {
+                case (_, Some(arn)) =>
+                  Redirect(routes.SubscriptionController.showSubscriptionComplete()).flashing("arn" -> arn.value)
```
johnrevill: I noticed there's a few metrics on the other outcomes in this endpoint - maybe we could add something like `mark("Count-Subscription-PartialSubscriptionCompleted")` to keep it consistent and complete?

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
...
+                                                                    case e => throw e
+                                                                  }
+                                                              case false =>
```
rvijayan: Looks like this case lead to showAlreadySubscribed in CheckAgencyController since we can get false from any SubscriptionState other than IsOnlySubscribedInETMP

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
...
+                                                                Future successful (false, None)
+                                                            }
+    } yield (wasPartiallySubscribed, optSucceededAllocatingArn)
+
+  def getSubscriptionStatus(utr: Utr, postcode: String)(
+    implicit hc: HeaderCarrier,
+    ec: ExecutionContext): Future[SubscriptionProcess] =
+    agentSubscriptionConnector.getRegistration(utr, postcode).map {
+
+      case Some(reg) if reg.isSubscribedToAgentServices && reg.isSubscribedToETMP =>
```
rvijayan: You can remove reg.isSubscribedToETMP from here since isSubscribedToAgentServices contains both.

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
arkowalski: this is different logic, it tells (wasPartiallySubscribed, optSucceededAllocatingArn) and you need Some(Arn) to be subscribed, this only happens if you are if !reg.isSubscribedToAgentServices && reg.isSubscribedToETMP
and your status is eligible for partialSubscription since SubscriptionState.IsOnlySubscribedInETMP,
you will land on showAlreadySubscribed if reg.isSubscribedToAgentServices && reg.isSubscribedToETMP which makes you SubscriptionState.IsSubscribedToAgentServices

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
...
   }
+
+  def completePartialSubscription(details: CompletePartialSubscriptionBody)(
+    implicit hc: HeaderCarrier,
+    ec: ExecutionContext): Future[(Boolean, Option[Arn])] =
+    for {
+      subscriptionProgress <- getSubscriptionStatus(details.utr, details.knownFacts.postcode) //FixMe APB-2805 this check should happen previously and can be stored in the session
+      isPartiallySubscribed = subscriptionProgress.state == SubscriptionState.IsOnlySubscribedInETMP
+      (wasPartiallySubscribed, optSucceededAllocatingArn) <- isPartiallySubscribed match {
+                                                              case true =>
```
johnrevill: Pattern matching on the boolean `isPartiallySubscribed` but could use `if(isPartiallySubscribed) {} else {}`

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
...
+      case Some(Registration(None, _, _)) =>
+        throw new IllegalStateException(s"The agency with UTR ${utr.value} has a missing organisation/individual name.")
+
+      case Some(reg) if !reg.isSubscribedToAgentServices && reg.isSubscribedToETMP =>
+        SubscriptionProcess(SubscriptionState.IsOnlySubscribedInETMP, Some(reg))
+
+      case Some(reg) if !reg.isSubscribedToAgentServices && !reg.isSubscribedToETMP =>
+        SubscriptionProcess(SubscriptionState.BrandNewSubscription, Some(reg))
+
+      case Some(reg) =>
```
rvijayan: I think you can remove this case since a (true, false) case is non existent.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
arkowalski: yup, done

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
arkowalski: true or false does not change much actually, you need Some(arn) to be successful.

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
arkowalski: ok, kept wonering whether ETMP record means anything if theyre already subscribed.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -72,8 +76,24 @@ class StartController @Inject()(
                   case Some(knownFacts) => sessionStoreService.cacheKnownFactsResult(knownFacts)
                   case None             => Future.successful(())
                 }
+            (wasPartiallySubscribed, optAllocatedArn) <- knownFactsResultOpt match {
```
johnrevill: There's a match on the `knownFactsResultOpt` on line 75. Could merge the match on line 79 into the match on line 75.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -72,8 +76,24 @@ class StartController @Inject()(
...
+                                                              CompletePartialSubscriptionBody(
+                                                                knownFact.utr,
+                                                                SubscriptionRequestKnownFacts(knownFact.postcode)))
+                                                          case None => Future successful (false, None)
+                                                        }
           } yield {
             knownFactsResultOpt match {
+              case Some(_) if wasPartiallySubscribed => {
+                val obtainedArn = optAllocatedArn
+                  .getOrElse(
```
johnrevill: That this case can even be considered might be symptomatic of a strange interface to the `completePartialSubscription` function.
I wonder if `completePartialSubscription` should just return an `Option[Arn]` - so if it's a Some then they were partially subscribed and now completely subscribed, and if it's a None they were and still are unsubscribed. 
That might help clear up some fiddly bits around calling the function and dealing with its result.

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SubscriptionService.scala
```diff
@@ -61,4 +82,50 @@ class SubscriptionService @Inject()(agentSubscriptionConnector: AgentSubscriptio
...
+      subscriptionProgress <- getSubscriptionStatus(details.utr, details.knownFacts.postcode) //FixMe APB-2805 this check should happen previously and can be stored in the session
+      isPartiallySubscribed = subscriptionProgress.state == SubscriptionState.IsOnlySubscribedInETMP
+      (wasPartiallySubscribed, optSucceededAllocatingArn) <- isPartiallySubscribed match {
+                                                              case true =>
+                                                                agentSubscriptionConnector
+                                                                  .completePartialSubscription(details)
+                                                                  .map(arn => (true, Some.apply(arn)))
+                                                                  .recover {
+                                                                    case e: Upstream4xxResponse
+                                                                        if Seq(Status.FORBIDDEN, Status.CONFLICT) contains e.upstreamResponseCode =>
```
johnrevill: Can we just allow the future to fail in this case?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyControllerISpec.scala
```diff
@@ -271,6 +272,46 @@ trait CheckAgencyControllerISpec extends BaseISpec with SessionDataMissingSpec {
...
+      AgentSubscriptionStub.partialSubscriptionWillSucceed(CompletePartialSubscriptionBody(validUtr,
+        knownFacts = SubscriptionRequestKnownFacts(validPostcode)))
+
+      implicit val request = authenticatedAs(subscribingCleanAgentWithoutEnrolments)
+        .withFormUrlEncodedBody("utr" -> validUtr.value, "postcode" -> validPostcode)
+      val result = await(controller.checkAgencyStatus(Some(CheckAgencyController.validBusinessTypes.head))(request))
+      redirectLocation(result) shouldBe Some(routes.SubscriptionController.showSubscriptionComplete().url)
+    }
+
+    "goGetCleanCreds, creds with enrolment/s are not allowed when partiallySubscribed User" in {
```
johnrevill: `goGetCleanCreds` - is that `showHasOtherEnrolments`?

arkowalski: yes I will rename to showHasOtherEnrolments

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
arkowalski: done

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
arkowalski: can you make a tech debt ticket for this

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CheckAgencyController.scala
johnrevill: I've just added case this as an acceptance critieria to the ticket.

PR #215, opened 2018-08-09
[APB-2605][JR,RV,AK] Auto-mapping of old GG account during subscription journey

johnrevill:


app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnector.scala
```diff
@@ -0,0 +1,73 @@
...
+    monitor(s"ConsumedAPI-Agent-Mapping-createPreSubscription-PUT") {
+      http
+        .PUT[String, HttpResponse](s"$mappingUrl/pre-subscription/${utr.value}", "")
+        .map(_ => ())
+        .recover {
+          case ex: Upstream4xxResponse if ex.upstreamResponseCode == 409 => ()
+        }
+    }
+
+  def updatePreSubscriptionWithArn(utr: Utr)(implicit hc: HeaderCarrier): Future[Unit] =
```
arkowalski: you should not need this unless it is execute the setup preSubscription mapping, and somehow we should capture whether the mapping was successful or not

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -39,7 +39,8 @@ class StartController @Inject()(
   val metrics: Metrics,
   override val appConfig: AppConfig,
   sessionStoreService: SessionStoreService,
-  subscriptionService: SubscriptionService)(implicit val aConfig: AppConfig)
+  subscriptionService: SubscriptionService,
+  subscriptionCtrlr: SubscriptionController)(implicit val aConfig: AppConfig)
```
arkowalski: you're importing a big controller here, rather not do that if we can. Something to discuss about in the workshop about design patterns. use reverse routing for this.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignedOutController.scala
```diff
@@ -38,8 +38,8 @@ class SignedOutController @Inject()(
     for {
       knownFactOpt <- sessionStoreService.fetchKnownFactsResult
       id <- knownFactOpt match {
-             case Some(x) => knownFactsResultMongoRepository.create(x).map(Option.apply)
-             case None    => Future successful None
+             case Some(knownFact) => knownFactsResultMongoRepository.create(knownFact).map(Option.apply)
```
rvijayan: You can avoid `map(Option.apply)` by mapping over knownFactOpt. That will remove the `case None` too

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -82,6 +85,14 @@ class SubscriptionController @Inject()(
 
   val desAddressForm = new DesAddressForm(Logger, blacklistedPostCodes)
 
+  private val linkAccountForm: Form[LinkAccount] =
```
arkowalski: we should have forms/models in models package
and optionally you rename "ans" & "lc" so its first glance readable, but its fine as is

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -187,33 +208,81 @@ class SubscriptionController @Inject()(
...
+            Ok(html.subscription_complete(continueUrl, isUrlToASAccount, wasEligibleForMapping, prettifiedArn))
+              .removingFromSession("arn")
+          }
         case _ =>
           Future.successful(sessionMissingRedirect())
       }
     }
   }
+
+  private def linkAccountResponse[A](
```
arkowalski: what happens if the mapping was requested, but failed? subscription would be complete at that point, leaving user confused. Just something to consider. 

also a lot of _ , perhaps for comprehension would make this clearer.

app/uk/gov/hmrc/agentsubscriptionfrontend/models/LinkAccount.scala
```diff
@@ -0,0 +1,37 @@
...
+
+package uk.gov.hmrc.agentsubscriptionfrontend.models
+
+sealed trait LinkAccountAnswer extends Product with Serializable
+
+object LinkAccountAnswer {
+  case object Yes extends LinkAccountAnswer
+  case object No extends LinkAccountAnswer
+
+  def apply(str: String): LinkAccountAnswer = str.toLowerCase match {
```
arkowalski: I assume the yes, no are always lower case when submitted using the form. So it should never fail, and if otherwise then should have tests assure correct values are submitted anyway.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -139,22 +150,32 @@ class SubscriptionController @Inject()(
...
         mark("Count-Subscription-AlreadySubscribed-APIResponse")
-        Redirect(routes.CheckAgencyController.showAlreadySubscribed())
+        Future.successful(Redirect(routes.CheckAgencyController.showAlreadySubscribed()))
 
       case Left(SubscriptionReturnedHttpError(status)) =>
         mark("Count-Subscription-Failed")
         throw new HttpException("Subscription failed", status)
     }
 
+  private[controllers] def redirectUponSuccessfulSubscription(arn: Arn)(implicit hc: HeaderCarrier) =
```
arkowalski: refactor this to what we have talked about, at this point the user would already be subscribed and not be informed about it, and they can drop out of the journey making changes to their account before ever receiving the complete page.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -82,6 +85,14 @@ class SubscriptionController @Inject()(
 
   val desAddressForm = new DesAddressForm(Logger, blacklistedPostCodes)
 
+  private val linkAccountForm: Form[LinkAccount] =
+    Form[LinkAccount](
+      mapping("autoMapping" -> optional(text).verifying(FieldMappings.radioInputSelected))(ans =>
```
rvijayan: May be we can avoid using optional(text) here as autoMapping always returns yes/no.

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnectorISpec.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  private implicit val hc = HeaderCarrier()
+
+  private lazy val connector: MappingConnector =
+    new MappingConnector(
+      new URL(s"http://localhost:$wireMockPort"),
+      app.injector.instanceOf[HttpGet with HttpPost with HttpPut with HttpDelete],
+      app.injector.instanceOf[Metrics])
+
+  private def withMetricsTimerUpdate(expectedMetricName: String)(testCode: => Unit): Unit = {
```
arkowalski: this would be useful in BaseISpec

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnectorISpec.scala
```diff
@@ -0,0 +1,164 @@
...
+    givenCleanMetricRegistry()
+    testCode
+    timerShouldExistAndBeUpdated(expectedMetricName)
+  }
+
+  "isEligible" when {
+    val withMetrics = withMetricsTimerUpdate("ConsumedAPI-Agent-Mapping-eligibility-GET") _
```
arkowalski: "_" here, do we need it?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -139,22 +150,32 @@ class SubscriptionController @Inject()(
...
     }
 
+  private[controllers] def redirectUponSuccessfulSubscription(arn: Arn)(implicit hc: HeaderCarrier) =
+    for (wasEligibileForMapping <- sessionStoreService.fetchMappingEligible) yield {
+      val redirectLocation = wasEligibileForMapping match {
+        case Some(true) => routes.SubscriptionController.showLinkAccount()
+        case _          => routes.SubscriptionController.showSubscriptionComplete()
+      }
+
+      Redirect(redirectLocation).withSession("arn" -> arn.value)
```
rvijayan: This will remove the existing sessions. You may need to use `withSession(request + ("arn" -> arn.value)`

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnector.scala
```diff
@@ -0,0 +1,73 @@
...
+class MappingConnector @Inject()(
+  @Named("agent-mapping-baseUrl") baseUrl: URL,
+  http: HttpGet with HttpPost with HttpPut with HttpDelete,
+  metrics: Metrics)
+    extends HttpAPIMonitor {
+  override val kenshooRegistry: MetricRegistry = metrics.defaultRegistry
+
+  private val mappingUrl = new URL(baseUrl, s"/agent-mapping/mappings")
+
+  def isEligibile(implicit hc: HeaderCarrier): Future[Boolean] =
```
arkowalski: we can improve on this, and just have the PUT endpoint, that would give us the answer to this, and if store them. Rather than checking and do it later anyway. I am sure the endpoint does these checks anyway, so its just the matter of returning valid messages and handling them, rather than making multiple calls.
Its more of always assuming successful journey, and if you get a response no, then its simply a no, otherwise you get success

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -187,33 +208,81 @@ class SubscriptionController @Inject()(
...
+  }
+
+  val submitLinkAccount: Action[AnyContent] = Action.async { implicit request =>
+    withAuthenticatedAgent {
+      request.session.get("arn").fold(Future.successful(sessionMissingRedirect())) { arn =>
+        linkAccountForm
+          .bindFromRequest()
+          .fold(
+            formWithErrors => {
+              if (formWithErrors.errors.exists(_.message == "error.link-account-value.invalid")) {
```
rvijayan: Not sure `error.link-account-value.invalid` has been added in the messages conf.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -187,33 +208,81 @@ class SubscriptionController @Inject()(
...
   val showSubscriptionComplete: Action[AnyContent] = Action.async { implicit request =>
+    def recoverSessionStoreWithNone[T]: PartialFunction[Throwable, Option[T]] = {
+      case NonFatal(ex) =>
+        Logger(getClass).warn("Session store service failure", ex)
+        None
+    }
+
     withAuthenticatedAgent {
-      request.flash.get("arn") match {
+      request.session.get("arn") match {
```
rvijayan: Looks like calling sessionMissingRedirect is repeated scenario when request.session.get("arn") is empty. We can have a common method for that. Can do it future..

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -73,26 +74,26 @@ class StartController @Inject()(
         } yield knownFactsResultOpt
       } else Future successful None
 
-      val decideOnSubscription = knownFactsResultOpt.flatMap {
+      knownFactsResultOpt.flatMap {
         case Some(knownFacts) =>
           for {
             _                   <- sessionStoreService.cacheKnownFactsResult(knownFacts)
             subscriptionProcess <- subscriptionService.getSubscriptionStatus(knownFacts.utr, knownFacts.postcode)
```
arkowalski: You have done what I have created ticket for before, and has been closed. Now that we are caching isSubscribedToETMP in the knownFacts. this is the flag for isPartiallySubscribed

arkowalski: subscriptionService.getSubscriptionStatus(knownFacts.utr, knownFacts.postcode)
This can be eliminated since we are caching the check for partially subscribed.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
arkowalski: I think we need the optional, just so we can return a valid error "no-radio-selected" if submitting nothing

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -84,90 +87,249 @@ class SubscriptionControllerISpec extends BaseISpec with SessionDataMissingSpec
...
+        "linkAccount.bullet-list.2",
+        "linkAccount.p3",
+        "linkAccount.p4",
+        "linkAccount.legend",
+        "linkAccount.option.yes",
+        "linkAccount.option.no"
+      )
+    }
+
+    "contain radio options for Yes and No" in new RequestAndResult {
```
rvijayan: Some unit tests for form validation would be nice...

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
```diff
@@ -84,90 +87,249 @@ class SubscriptionControllerISpec extends BaseISpec with SessionDataMissingSpec
...
+      "return 200 and redisplay the /link-account page with an error message for missing choice" in {
+        val request = authenticatedAs(subscribingAgentEnrolledForHMRCASAGENT)
+          .withSession("arn" -> "AARN0000001")
+
+        resultOf(request) should containMessages("linkAccount.title", "error.no-radio-selected")
+      }
+    }
+
+    "form value is invalid" should {
+      "result in a BadRequest" in new RequestWithSessionDetails(autoMappingFormValue = "somethingInvalid") {
```
rvijayan: I think it will return a match error which may result in internal server error other than bad request. We could throw a bad-request-exception for invalid values of that form.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
johnrevill: Yes, I wasn't happy with this either, controllers pulling in other controllers as dependencies. 
The only thing it uses is the `SubscriptionController.redirectUponSuccessfulSubscription` funnction which is re-used in 3 different places in two different controllers and at least keeps in one place all the routing for whether the `/link-account` page should be shown or not.
It should live somewhere else.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: For future tech debt/discussion.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: So I think what you're getting at is that the subscription is made before the '/link-account' page is shown rather than afterwards.
If a mapping call fails after they click Yes to the '/link-account' page, they'd see the usual "There was a problem page, please retry again later". If they retried again later, they'd just be sent straight to the AS homepage and the mapping may or may not have been updated (if not, it'd be cleaned up by the mongo TTL).
If we instead delayed the subscription until after the `/link-account` page, then the mapping calls fail, they'd still see the same "There was a problem page, please retry again later" page and be in the exact same situation.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
arkowalski: reverse routing is fine, you're only redirecting, so there is no need to import this SubscriptionController

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnector.scala
arkowalski: Yes, I know understand what this does. However we should make this call do more than it already does.
Additionally this should be our check whether they have eligableEnrolments to map, and if theyre are any successfully create an UPSTREAM preSubscription.
Else return a valid answer with No eligible enrolments

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignedOutController.scala
```diff
@@ -30,16 +32,26 @@ import scala.concurrent.Future
...
-             case None            => Future successful None
+             case Some(knownFact) => {
+               for {
+                 isEligibleForMapping <- mappingConnector.isEligibile
+                 _ <- if (isEligibleForMapping) mappingConnector.createPreSubscription(knownFact.utr)
+                     else Future.successful(())
+                 idOpt <- chainedSessionRepository
+                           .create(ChainedSessionDetails(knownFact, isEligibleForMapping))
+                           .map(Option.apply)
+               } yield idOpt
```
arkowalski: this ID should never be an option. Is this possible?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
```diff
@@ -107,89 +107,89 @@ class StartControllerISpec extends BaseISpec {
...
       }
     }
 
-    "redirect to /subscription-complete if given a valid KnownFactsResult ID and agent is partially subscribed (subscribed in ETMP but not enrolled)" in new ValidKnownFactsCached {
-      AgentSubscriptionStub.withMatchingUtrAndPostcode(
-        knownFactsResult.utr,
-        knownFactsResult.postcode,
-        isSubscribedToAgentServices = false,
-        isSubscribedToETMP = true)
+    "redirect to correct page if given a valid KnownFactsResult ID and agent is partially subscribed (subscribed in ETMP but not enrolled)" when {
```
arkowalski: not a KnownFactsResult

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
```diff
@@ -212,21 +212,29 @@ class StartControllerISpec extends BaseISpec {
       redirectLocation(result).head should include(routes.CheckAgencyController.showCheckBusinessType().url)
     }
 
-    "delete the persisted KnownFactsResult if given a valid KnownFactsResult ID" in new ValidKnownFactsCached {
+    "delete the persisted KnownFactsResult if given a valid KnownFactsResult ID" in new ValidKnownFactsCached with UnsubscribedAgentStub {
```
arkowalski: test name again

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
```diff
@@ -212,21 +212,29 @@ class StartControllerISpec extends BaseISpec {
...
       await(controller.returnAfterGGCredsCreated(id = Some(persistedId))(FakeRequest()))
 
-      await(repo.findKnownFactsResult(persistedId)) shouldBe None
+      await(repo.findChainedSessionDetails(persistedId)) shouldBe None
     }
 
-    "repopulate the KnownFacts session store with the persisted KnownFactsResult, if given a valid KnownFactsResult ID" in new ValidKnownFactsCached {
+    "place the known facts back in the session store, if given a valid ChainedSessionDetails ID" in new ValidKnownFactsCached with UnsubscribedAgentStub {
```
arkowalski: its not just the known facts, you're storing the mapping boolean as well

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnectorISpec.scala
johnrevill: Done. Pulled it down into BaseISpec and updating all tests that check metrics to use it

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnectorISpec.scala
johnrevill: Yep `withMetrics` is a curried function of the `withMetricsTimerUpdate` function

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: Nice spot. Updated, thanks!

app/uk/gov/hmrc/agentsubscriptionfrontend/repository/ChainedSessionDetailsRepository.scala
```diff
@@ -18,26 +18,26 @@ package uk.gov.hmrc.agentsubscriptionfrontend.repository
...
+import uk.gov.hmrc.agentsubscriptionfrontend.repository.StashedChainedSessionDetails.StashedChainnedSessionId
 import uk.gov.hmrc.mongo.ReactiveRepository
 import uk.gov.hmrc.mongo.json.ReactiveMongoFormats
 
 import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
-class KnownFactsResultMongoRepository @Inject()(appConfig: AppConfig, mongoComponent: ReactiveMongoComponent)
-    extends ReactiveRepository[StashedKnownFactsResult, BSONObjectID](
-      "agent-known-facts-results",
```
arkowalski: (collection not used any more)if this can be cleaned up easily, lets do it

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: This is the case where the user has been sending up values in the request that aren't even possible from the rendered HTML form.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
arkowalski: lets move the /link-account page before subscription happens. So if they ever drop out at that page, their account is unaffected. They must complete the journey for any changes to take effect to their account.
so, they must HIT the confirm details button.

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnectorISpec.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    "agent-mapping returns a 200 response" should {
+      "return successful future" in withMetrics {
+        MappingStubs.givenMappingUpdateToPostSubscription(Utr("1234567890"), httpReturnCode = 200)
+        await(connector.updatePreSubscriptionWithArn(Utr("1234567890")))
+        MappingStubs.verifyMappingUpdateToPostSubscriptionCalled(Utr("1234567890"))
+      }
+    }
+
+    "fail with Upstream4xxException if agent-mapping fails with a 401 response" in withMetrics {
```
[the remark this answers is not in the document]
johnrevill: Have updated the test names & structure to all be consistent in this ISpec.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: For future tech debt/discussion

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
johnrevill: These are the tests for the GET request so are just checking the rendered HTML has the expected content. The checks for the form validation are in the tests for the POST endpoint below

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AddressLookupFrontendConnectorSpec.scala
```diff
@@ -13,23 +13,21 @@ class AddressLookupFrontendConnectorSpec extends BaseISpec with MetricTestSuppor
...
-        country = Country("GB", Some("United Kingdom")))
-
-      timerShouldExistAndBeUpdated("ConsumedAPI-Address-Lookup-Frontend-getAddressDetails-GET")
+      withMetricsTimerUpdate("ConsumedAPI-Address-Lookup-Frontend-getAddressDetails-GET") {
+        val addressId = "id"
+        val addressLine1 = "10 Other Place"
+        val addressLine2 = "Some District"
+        val addressLine3 = "Line 3"
+        val town = "Our town"
+        val postcode = "AA1 1AA"
```
arkowalski: in code, we should use normalised style of data. so all inputs should be evaluated after form submission, otherwise there is no spacing.
if you're using this in tests, then modify the postcode in that particular test, but I see this is supposed to be a valid postcode so keep it 100% valid

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionControllerISpec.scala
johnrevill: Done

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignedOutController.scala
johnrevill: Nice, thanks

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartControllerISpec.scala
johnrevill: Yep, the next test covers just that

app/uk/gov/hmrc/agentsubscriptionfrontend/repository/ChainedSessionDetailsRepository.scala
johnrevill: Future tech debt/discussion

it/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AddressLookupFrontendConnectorSpec.scala
johnrevill: It's all good. 
The DES compliant validation is performed upon returning from the address lookup in the `SubscriptionController.returnFromAddressLookup` endpoint which is covered in the `SubscriptionControllerISpec` tests.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: Will raise a tech debt ticket for this so we don't forget and can discuss it

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: Will raise a tech debt ticket for this so we don't forget and can discuss it

johnrevill: Captured in APB-2891

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: Captured in APB-2891

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
johnrevill: Captured in tech debt ticket ASA-69

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/MappingConnector.scala
johnrevill: Captured in tech debt ticket ASA-70

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
johnrevill: Captured in tech debt ticket ASA-71

PR #228, opened 2018-08-22
[ASA-64][AK,JR] auto mapping feature flag

arkowalski:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -286,36 +293,44 @@ class SubscriptionController @Inject()(
...
+      case true =>
+        withAuthenticatedAgent {
+          Future.successful {
+            request.session.get("arn").fold(sessionMissingRedirect()) { _ =>
+              Ok(html.link_account(linkAccountForm))
+            }
+          }
         }
-      }
+      case false => Future.successful(BadRequest)
```
rvijayan: I would suggest to use 500 exception since the its us causing the exception, so that the end user will will get "there is something wrong" page.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -286,36 +293,44 @@ class SubscriptionController @Inject()(
...
+                validatedLinkAccount => {
+                  validatedLinkAccount.autoMapping match {
+                    case Yes => linkAccountResponse(mappingConnector.updatePreSubscriptionWithArn)
+                    case No  => linkAccountResponse(mappingConnector.deletePreSubscription)
+                  }
+                }
+              )
+          }
+        }
+      case false => Future.successful(BadRequest)
```
rvijayan: 500 exception is more suitable here

PR #237, opened 2018-09-04
[JR] Fix propagation of ARN through to /subscription-complete for partially subscribed agents

johnrevill:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessIdentificationController.scala
```diff
@@ -173,8 +173,10 @@ class BusinessIdentificationController @Inject()(
             .fold(
               formWithErrors => Future successful Ok(html.business_details(formWithErrors, businessTypeIdentifier)),
               knownFacts =>
-                checkBusinessDetailsGivenValidForm(knownFacts).map(
-                  _.withSession(request.session + ("businessType" -> businessTypeIdentifier)))
+                checkBusinessDetailsGivenValidForm(knownFacts).map { resultWithSession =>
+                  val sessionData = (request.session.data ++ resultWithSession.session.data.toSeq) + ("businessType" -> businessTypeIdentifier)
```
rvijayan: Should checkBusinessDetailsGivenValidForm return an ARN and we can do the redirect here, in the submitBusinessDetailsForm action? That way, we can avoid the sessionData value and keep Arn and businessType in the session.

rvijayan: Ignore the above comment. checkBusinessDetailsGivenValidForm is dealing with many redirects :(

PR #251, opened 2018-09-21
[ASA-124][JR] Supporting old start endpoints: /check-agency-status and /check-business-type. They both capture any continue URL and redirect to /business-type

johnrevill:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessIdentificationController.scala
```diff
@@ -100,17 +108,14 @@ class BusinessIdentificationController @Inject()(
 
   def showBusinessDetailsForm(businessType: Option[String]): Action[AnyContent] = Action.async { implicit request =>
     withSubscribingAgent { implicit agent =>
-      //withMaybeContinueUrlCached because, Currently still needed as a user might be arriving from: trusts registration flow or gov.uk guidence page, make sure this is not the case anymore before removing
-      withMaybeContinueUrlCached {
```
rvijayan: Should we keep this withMaybeContinueUrlCached as it is, since we are declaring the check-agency-status and check-business-type endpoints are deprecated. That means, users can use this action directly.

PR #253, opened 2018-09-24
[ASA-49 & 72][AK] showLinkClients before subscriptionComplete page, fetches arn after subscription to display rather than store in session

arkowalski:


app/uk/gov/hmrc/agentsubscriptionfrontend/auth/AuthActions.scala
```diff
@@ -61,6 +62,26 @@ trait AuthActions extends AuthorisedFunctions with AuthRedirects with Monitoring
...
+    authorised(Enrolment("HMRC-AS-AGENT") and AuthProviders(GovernmentGateway))
+      .retrieve(authorisedEnrolments) {
+        case enrolments =>
+          body(
+            getEnrolmentValue(enrolments, "HMRC-AS-AGENT", "AgentReferenceNumber")
+              .map(Arn(_))
+              .getOrElse(throw new InsufficientEnrolments("could not find the HMRC-AS-AGENT enrolment to continue")))
+      }
+      .recover {
+        case _: UnsupportedAffinityGroup =>
```
rvijayan: Not sure this is a case anymore. But I think `authorised(Enrolment("HMRC-AS-AGENT")...` will throw an InsufficientEnrolments exception. We may need to handle that.

app/uk/gov/hmrc/agentsubscriptionfrontend/auth/AuthActions.scala
```diff
@@ -61,6 +62,26 @@ trait AuthActions extends AuthorisedFunctions with AuthRedirects with Monitoring
...
+          mark("Count-Subscription-NonAgent")
+          Redirect(routes.StartController.showNotAgent())
+        case _: NoActiveSession =>
```
rvijayan: I think ErrorHandler will handle this case well with an audit.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CommonRouting.scala
```diff
@@ -15,35 +15,75 @@
...
+  private[controllers] def completeMappingWhenAvailable(utr: Utr, completedPartialSub: Boolean = false)(
+    implicit request: Request[AnyContent],
+    hc: HeaderCarrier,
+    ec: ExecutionContext): Future[Result] = {
+
+    val doMappingAnswer: Boolean = request.session.get("performAutoMapping").isDefined || completedPartialSub
+
+    for {
+      completeMapping <- if (appConfig.autoMapAgentEnrolments && doMappingAnswer)
+                          mappingConnector.updatePreSubscriptionWithArn(utr)
```
rvijayan: I dont think we should perform some logic other than routing here.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CommonRouting.scala
```diff
@@ -15,35 +15,75 @@
...
+                        else Future successful ()
+    } yield Redirect(routes.SubscriptionController.showSubscriptionComplete())
+  }
+
+  private[controllers] def handlePartialSubscription(
```
rvijayan: Not sure this method belong to CommonRouting class.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CommonRouting.scala
```diff
@@ -15,35 +15,75 @@
...
+  subscriptionService: SubscriptionService,
+  appConfig: AppConfig)
+    extends Monitoring {
+
+  private[controllers] def completeMappingWhenAvailable(utr: Utr, completedPartialSub: Boolean = false)(
+    implicit request: Request[AnyContent],
+    hc: HeaderCarrier,
+    ec: ExecutionContext): Future[Result] = {
+
+    val doMappingAnswer: Boolean = request.session.get("performAutoMapping").isDefined || completedPartialSub
```
rvijayan: I would prefer to keep the usage of session out of the method. It will improve the reusability of the method. I can see completeMappingWhenAvailable is used in redirectSubscriptionResponse which may not have a clue about that session.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -217,27 +235,19 @@ class SubscriptionController @Inject()(
...
+        continueUrlOpt           <- sessionStoreService.fetchContinueUrl.recover(recoverSessionStoreWithNone)
+        wasEligibleForMappingOpt <- sessionStoreService.fetchMappingEligible.recover(recoverSessionStoreWithNone)
+        _                        <- sessionStoreService.remove()
+      } yield {
+        val continueUrl = continueUrlOpt.map(_.url).getOrElse(appConfig.agentServicesAccountUrl)
+        val isUrlToASAccount = continueUrlOpt.isEmpty
+        val wasEligibleForMapping = wasEligibleForMappingOpt.contains(true)
+        val prettifiedArn = TaxIdentifierFormatters.prettify(arn)
+        Ok(html.subscription_complete(continueUrl, isUrlToASAccount, wasEligibleForMapping, prettifiedArn))
+          .removingFromSession("arn")
```
rvijayan: I think we are not keeping arn in session

app/uk/gov/hmrc/agentsubscriptionfrontend/views/link_clients.scala.html
```diff
@@ -21,27 +21,30 @@
...
-    </ul>
-
-    <p>@Messages("linkClients.p3")</p>
 
-    <p>@Messages("linkClients.p4")</p>
+    <div class="form-group">
+        <details role="group">
+            <summary role="button" aria-controls="details-content-99ee20e1-c513-460b-83d1-2311b314d0b9" aria-expanded="false">
+                <span class="summary">@Messages("linkClients.expand.button")</span></summary>
+            <div class="panel panel-border-narrow" id="details-content-99ee20e1-c513-460b-83d1-2311b314d0b9" aria-hidden="true">
```
rvijayan: Why does this id contain digits -99ee20e1-c513-460b-83d1-2311b314d0b9?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/CommonRouting.scala
arkowalski: obtaining the performAutoMapping in session is the step before redirectSubscriptionResponse, and it is optional as it accepts the completedPartialSub parameter. essentially you will either be able to activate this method through session or available parameter. We need to store it in the session, as it is switching pages and its not all done in one step... Whereas when its called from the other place, it is capable of executing the function directly, hence why not re-using the session.

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -198,14 +197,32 @@ class SubscriptionController @Inject()(
...
                 validatedLinkClients => {
+                  val isPartiallySubscribed = request.session.get("isPartiallySubscribed").isDefined
+
                   validatedLinkClients.autoMapping match {
-                    case Yes => linkClientsResponse(mappingConnector.updatePreSubscriptionWithArn)
-                    case No  => linkClientsResponse(mappingConnector.deletePreSubscription)
+                    case Yes =>
+                      if (isPartiallySubscribed)
+                        for {
+                          completePartialSub <- subscriptionService.completePartialSubscription(
```
rvijayan: This param - completePartialSub is never used?

PR #254, opened 2018-10-01
[ASA-195][SS] accessibility fixes

somme:


app/assets/stylesheets/styles.css
```diff
@@ -120,3 +120,31 @@
...
+    margin-bottom: 30px;
+    padding: 20px 15px 15px;
+}
+.flash.error-summary a{
+    font-weight: bold;
+}
+#error-summary-heading{
+    font-size: 24px;
+}
+.form-field--error .error-notification{
```
arkowalski: .form-field--error is not supposed to be here? I see it below

PR #280, opened 2019-03-11
ASA-802[AL] Update error pages - 403,404,500

adamliversidge:


app/ErrorHandler.scala
```diff
@@ -55,11 +55,15 @@ class ErrorHandler @Inject()(
...
         toGGLogin(if (env.mode.equals(Mode.Dev)) s"http://${request.host}${request.uri}" else s"${request.uri}")
-      case _: InsufficientEnrolments => Forbidden
-      case _                         => super.resolveError(request, exception)
+      case _: InsufficientEnrolments =>
+        Forbidden(
+          standardErrorTemplate("global.error.403.title", "global.error.403.heading", "global.error.403.message")(
+            request))
+      case _ => super.resolveError(request, exception)
     }
   }
```
geoffreywatson: Is there a test in `ViewSpec` we can write to support this?

PR #282, opened 2019-03-14
[ASA-620][SM]: UK Sub - build AA check: CT UTR

sureshhmrc:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessIdentificationController.scala
```diff
@@ -102,13 +102,47 @@ class BusinessIdentificationController @Inject()(
...
             else
-              Redirect(
-                routes.BusinessIdentificationController.showBusinessDetailsForm(validatedBusinessType.businessType))
+              Redirect(routes.BusinessIdentificationController.showUtrForm(validatedBusinessType.businessType))
           }
         )
     }
   }
 
+  def showUtrForm(businessType: IdentifyBusinessType): Action[AnyContent] = Action.async { implicit request =>
```
sureshhmrc: passing `businessType` across endpoints back and forth is bas  - instead we should store it in the session and use it every where we need  -  but thats how existing code is written - needs to refactor the cache handling

app/uk/gov/hmrc/agentsubscriptionfrontend/service/SessionStoreService.scala
```diff
@@ -72,6 +72,12 @@ class SessionStoreService @Inject()(sessionCache: SessionCache) {
   def fetchIsChangingAnswers(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Option[Boolean]] =
     sessionCache.fetchAndGetEntry[Boolean]("isChangingAnswers")
 
+  def cacheAgentSession(agentSession: AgentSession)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Unit] =
+    sessionCache.cache("agentSession", agentSession).map(_ => ())
+
+  def fetchAgentSession(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Option[AgentSession]] =
+    sessionCache.fetchAndGetEntry[AgentSession]("agentSession")
```
sureshhmrc: we should have a global case class for holding all data - we can not keep on adding stuff like this -

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessIdentificationController.scala
geoffreywatson: Sounds good I would agree storing it the session would be better.

PR #303, opened 2019-05-28
[ASA-1007][GW] remove invalid radio option from business type

geoffreywatson:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessTypeController.scala
```diff
@@ -64,24 +63,15 @@ class BusinessTypeController @Inject()(
         .fold(
           formWithErrors => Ok(html.business_type(formWithErrors)),
           validatedBusinessType => {
-            if (validatedBusinessType == Invalid)
```
sureshhmrc: would this lead to technical error if user manipulates the form and passes some random string to businessType and submits the form ?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessTypeControllerISpec.scala
```diff
@@ -85,37 +84,6 @@ class BusinessTypeControllerISpec extends BaseISpec with SessionDataMissingSpec
...
-    "display invalid business type page" in {
-      val request = authenticatedAs(subscribingAgentEnrolledForNonMTD)
-
-      val result = await(controller.showInvalidBusinessType(request))
-
-      result should containMessages("invalid.businessType.title",
-        "invalid.businessType.p1", "invalid.businessType.l1",
-        "invalid.businessType.l2", "invalid.businessType.l3",
-        "invalid.businessType.l4", "invalid.businessType.button")
-    }
```
sureshhmrc: we better have a test for checking the invalid business type form

PR #329, opened 2019-06-21
[APB-3768][MP] New task list page with identify business page function completed

mariannepearson0:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListController.scala
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.agentsubscriptionfrontend.controllers
+import com.kenshoo.play.metrics.Metrics
+import javax.inject.Inject
+import play.api.i18n.MessagesApi
+import play.api.mvc.{Action, AnyContent}
+import uk.gov.hmrc.agentsubscriptionfrontend.config.AppConfig
+import uk.gov.hmrc.agentsubscriptionfrontend.service.{SessionStoreService, SubscriptionService}
```
sureshhmrc: unused import ?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListController.scala
```diff
@@ -0,0 +1,47 @@
...
+  implicit override implicit val appConfig: AppConfig,
+  metrics: Metrics,
+  override val messagesApi: MessagesApi,
+  val ec: ExecutionContext)
+    extends AgentSubscriptionBaseController(authConnector, continueUrlActions, appConfig) with SessionBehaviour {
+
+  def showTaskList: Action[AnyContent] = Action.async { implicit request =>
+    withSubscribingAgent { implicit agent =>
+      sessionStoreService.fetchAgentSession.map {
+        case Some(session) => Ok(html.task_list(session.identifyBusinessTaskComplete))
```
sureshhmrc: having a valid session doesn't necessarily mean that the business details task is done , we need to check if valid business details exist in the session or not ,
or am I missing something ?

app/uk/gov/hmrc/agentsubscriptionfrontend/models/AgentSession.scala
```diff
@@ -32,7 +32,9 @@ case class AgentSession(
   registration: Option[Registration] = None,
   checkAmls: Option[String] = None,
   amlsAppliedFor: Option[String] = None,
-  amlsDetails: Option[AMLSDetails] = None)
+  amlsDetails: Option[AMLSDetails] = None,
+  identifyBusinessTaskComplete: Boolean = false,
```
sureshhmrc: can we use short name here: `businessTaskComplete`

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListControllerISpec.scala
```diff
@@ -0,0 +1,33 @@
...
+      result should containMessages(
+        "task-list.header",
+        "task-list.1.header",
+        "task-list.2.header",
+        "task-list.3.header",
+        "task-list.4.header")
+    }
+    "contain CONTINUE tag when a task has been completed" in {
+      implicit val request = authenticatedAs(subscribingAgentEnrolledForNonMTD)
+      sessionStoreService.currentSession.agentSession = Some(AgentSession(identifyBusinessTaskComplete = true))
```
sureshhmrc: we need to add a test where we have a session but no business details exists, then verify if we actually show the flag as true or false

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessIdentificationController.scala
```diff
@@ -146,16 +149,22 @@ class BusinessIdentificationController @Inject()(
...
+    implicit hc: HeaderCarrier): Future[Call] =
     businessDetailsValidator.validate(existingSession.registration) match {
       case Failure(responses) if responses.contains(InvalidBusinessName) =>
         routes.BusinessIdentificationController.showBusinessNameForm()
       case Failure(responses) if responses.exists(r => r == InvalidBusinessAddress || r == DisallowedPostcode) =>
         routes.BusinessIdentificationController.showUpdateBusinessAddressForm()
       case Failure(responses) if responses.contains(InvalidEmail) =>
         routes.BusinessIdentificationController.showBusinessEmailForm()
-      case _ =>
+      case _ if continueUrl.isDefined =>
```
sureshhmrc: I guess this fails for users who login with unclean creds ( ? ) for users with unclean creds there will be a continueUrl as far as I remember , but please double check that we are not breaking the journey ...

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListController.scala
mariannepearson0: 👍

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListController.scala
mariannepearson0: there's a flag in the session that gets set to true at the end of the business details section (after you confirm the business). If the business task isn't done the task list will still be shown but with only the business details link clickable so they can complete this section.

app/uk/gov/hmrc/agentsubscriptionfrontend/models/AgentSession.scala
mariannepearson0: sure

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/BusinessIdentificationController.scala
mariannepearson0: its for users coming from Trusts. John informed me that only users coming from Trusts will have a continue url and we don't want them to see the task list because they haven't paid for it or some other reasons Ian gave me.

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListControllerISpec.scala
mariannepearson0: So the flag is set upon the submission of /confirm-business back to the track page, it's not based on the business details in the session

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/TaskListController.scala
sureshhmrc: ah yes, i misread the above logic , i thought you are simply doing `.task_list(identifyBusinessTaskComplete)` but you are actually doing `.task_list(session.identifyBusinessTaskComplete)` 

but don't we need to check for the continueUrl here too as if someone manually hits the  /task-list url (from Trusts) then the task list page will be shown , right ?

mariannepearson0: yeah thats true, ok!

PR #336, opened 2019-06-28
[APB-3770][MP] fix to allow agents with clean creds to skip create Id task

mariannepearson0:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/AMLSController.scala
```diff
@@ -286,4 +276,21 @@ class AMLSController @Inject()(
         Redirect(routes.BusinessDetailsController.showBusinessDetailsForm())
       //Redirect(routes.UtrController.showUtrForm())
     }
+
+  def updateSession(existingSession: AgentSession, amlsDetails: AMLSDetails, agent: Agent)(
+    implicit hc: HeaderCarrier) = {
+    val newSession = agent match {
+      case hasNonEmptyEnrolments(_) =>
```
JTY009: hasEnrolments?

mariannepearson0: yeah i didn't hame this extractor but I think it makes sense because you don't really care what the enrolments are just that there are some (or not none :-) ). And the actual implementation is just checking that the enrolments are non empty so it does what it says on the tin

PR #368, opened 2019-07-26
APB-3900 wire up buttons on AMLS pages

geoffreywatson:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/AMLSController.scala
```diff
@@ -250,15 +256,19 @@ class AMLSController @Inject()(
                 val supervisoryBodyData =
                   amlsBodies.getOrElse(validForm.amlsCode, throw new Exception("Invalid AMLS code"))
 
+                val continue =
```
stevenmallen: This expression looks the same as the one above - can we refactor to a common method?

PR #375, opened 2019-07-29
Enable Wart remover; fix (most of) the errors it found

stevenmallen:
Mostly the following needed to be changed:
* Replace Enumerations with sealed abstract classes
* use JavaConverters rather than JavaConversions
* Remove (most) Option.get usages
* Use explicit type annotation for implicit vals
* Remove usages of null

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/models/AMLSDetails.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/models/AMLSDetails.scala
@@ -24,13 +24,13 @@ import play.api.libs.json._
 case class RegisteredDetails(membershipNumber: String, membershipExpiresOn: LocalDate)
 
 object RegisteredDetails {
-  implicit val format = Json.format[RegisteredDetails]
+  implicit val format: OFormat[RegisteredDetails] = Json.format[RegisteredDetails]
 }
 
 case class PendingDetails(appliedOn: LocalDate)
 
 object PendingDetails {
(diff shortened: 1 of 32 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/models/AMLSDetails.scala
```diff
@@ -24,13 +24,13 @@ import play.api.libs.json._
 case class RegisteredDetails(membershipNumber: String, membershipExpiresOn: LocalDate)
 
 object RegisteredDetails {
-  implicit val format = Json.format[RegisteredDetails]
+  implicit val format: OFormat[RegisteredDetails] = Json.format[RegisteredDetails]
```
mariannepearson0: could remove the type after Json.format now that type is explicitly annotated

app/uk/gov/hmrc/agentsubscriptionfrontend/views/govuk_wrapper.scala.html
```diff
@@ -98,8 +98,9 @@
 }
 
 @mainContentHeader = {
-    @if(contentHeader.isDefined) {
-        @uiLayouts.main_content_header(contentHeader = contentHeader.get)
+    @contentHeader match {
+      case Some(header) => { @uiLayouts.main_content_header(contentHeader = header) }
+      case None => { }
```
mariannepearson0: would maybe work better as a fold, as we don't really care about the None case

PR #393, opened 2019-08-07
[APB-4095][MP] Rework partial subscription scenarios now record is deleted after ETMP record creation

mariannepearson0:
Now that the subscription journey record is deleted upon successful creation of a record in ETMP, there is no way for a partially subscribed user to come into the journey with a subscription journey record. So we can check when the user enters their UTR and postcode if they are partially subscribed then if they are a clean creds user complete the subscription and if they are unclean creds ask them to sign in again with the clean creds they have previously created.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
@@ -120,15 +120,12 @@ class SubscriptionController @Inject()(
         val sjr = agent.getMandatorySubscriptionRecord
         (sjr.businessDetails.utr, sjr.businessDetails.postcode, sjr.businessDetails.registration, sjr.amlsData) match {
           case (utr, postcode, Some(registration), amlsData) =>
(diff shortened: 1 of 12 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SubscriptionController.scala
```diff
@@ -101,16 +101,31 @@ class SubscriptionController @Inject()(
...
     withSubscribingAgent { agent =>
       agent.withCleanCredsOrSignIn {
         val sjr = agent.getMandatorySubscriptionRecord
         (sjr.businessDetails.utr, sjr.businessDetails.postcode, sjr.businessDetails.registration, sjr.amlsData) match {
           case (utr, postcode, Some(registration), amlsData) =>
-            subscriptionService.handlePartiallySubscribedAndRedirect(agent, sjr.businessDetails.utr, sjr.businessDetails.postcode)(
-              whenNotPartiallySubscribed = subscriptionService
-              .subscribe(utr, postcode, registration, amlsData)
-              .flatMap(redirectSubscriptionResponse(_, agent)))
+                    for {
```
stevenmallen: Could you please add a comment about why we need to update the session - (for the subscription complete page, after the record is deleted)

PR #396, opened 2019-08-09
[APB-4051][MP] using auth id to retrieve sjr when returning from mapping instead of continue id

mariannepearson0:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -88,17 +88,15 @@ class StartController @Inject()(
...
       continueUrlActions.withMaybeContinueUrlCached {
-        id match {
-          case Some(continueId) =>
-            for {
-              record <- subscriptionJourneyService.getMandatoryJourneyRecord(ContinueId(continueId))
-              _      <- subscriptionJourneyService.saveJourneyRecord(record.copy(mappingComplete = true))
-            } yield Redirect(routes.TaskListController.showTaskList())
-
-          case None => Future.successful(Redirect(routes.TaskListController.showTaskList()))
+        sjr.authProviderId flatMap { authProviderId =>
```
stevenmallen: Trying to figure out why you need to flatMap here (I guess it's a Future, but it's weird because sjr is not a future)?  But it compiles, so never mind :)

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -88,17 +88,15 @@ class StartController @Inject()(
...
+          for {
+            record <- subscriptionJourneyService.getMandatoryJourneyRecord(authProviderId)
```
stevenmallen: Not sure why you can't update the sjr you got above?  Why do you need to refetch it again?

stevenmallen: would this work too:

```
def returnAfterMapping(): Action[AnyContent] = Action.async { implicit request =>
    withSubscribingAgent { agent =>
      continueUrlActions.withMaybeContinueUrlCached {
        val sjr = agent.getMandatorySubscriptionRecord
        subscriptionJourneyService.saveJourneyRecord(sjr.copy(mappingComplete = true))
          .map(_ => Redirect(routes.TaskListController.showTaskList()))
      }
    }
  }
```

mariannepearson0: yeah sorry, too excited about lunch, will fix now

PR #415, opened 2019-09-09
[APB-4094][MP] Convert deprecated ContinueUrl to RedirectUrl

mariannepearson0:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/RedirectUrlActions.scala
```diff
@@ -0,0 +1,91 @@
...
+import uk.gov.hmrc.agentsubscriptionfrontend.connectors.SsoConnector
+import uk.gov.hmrc.agentsubscriptionfrontend.service.SessionStoreService
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.bootstrap.binders.RedirectUrl._
+import uk.gov.hmrc.play.bootstrap.binders.{AbsoluteWithHostnameFromWhitelist, RedirectUrl, UnsafePermitAll}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.language.higherKinds
+import scala.util.{Failure, Success, Try}
+
```
stevenmallen: Hey, I know comments are seen as an anti-pattern, but it would be really nice to have a simple comment block at the top of this file which describes exactly what a "RedirectUrl" is in this context.  People might think it's just a standard HTTP 303, but I think the intention here is something different?  This is regarding what happens at the end of a journey, right?  Since RedirectUrl is such a general term, I think a comment would be helpful :)

mariannepearson0: 👍

conf/application.conf
```diff
@@ -123,7 +123,6 @@ Dev {
       sso {
         host = localhost
         port = 9974
-        protocol = https
```
stevenmallen: Why was this removed?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/SignOutControllerISpec.scala
```diff
@@ -51,17 +51,17 @@ class SignOutControllerISpec extends BaseISpec {
         s"continue=%2Fagent-subscription%2Freturn-after-gg-creds-created%3Fid%3Dfoo")
     }
 
-    def assertContinueUrl(result: Result, continueUrl: ContinueUrl): Assertion = {
+    def assertContinueUrl(result: Result, continueUrl: String): Assertion = {
       val sosContinueValueUnencoded =
-        s"/agent-subscription/return-after-gg-creds-created?continue=${continueUrl.encodedUrl}"
+        s"/agent-subscription/return-after-gg-creds-created?continue=${URLEncoder.encode(continueUrl, "UTF-8")}"
```
stevenmallen: Thank you for spotting and fixing this :)

conf/application.conf
mariannepearson0: when running it locally I was getting a no SSL certificate error (or something like this) when this endpoint was called because It should be http protocol in local...? http is the default if no protocol is provided

PR #425, opened 2019-09-19
[APB-4361][MP] Move trigger points to complete partial subscription

mariannepearson0:
move the completion of partial subscription after check answers instead of after postcode and enable ability to complete subscription after returning from creating a new user ID. This is related to the live issue in ticket APB-4361.

### DIFF
--- a/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
+++ b/app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
@@ -75,28 +75,7 @@ class StartController @Inject()(
             // sanity check - they just came back with a brand new Auth Id
             require(agent.subscriptionJourneyRecord.isEmpty)
 
-            for {
-              record <- subscriptionJourneyService.getMandatoryJourneyRecord(ContinueId(continueId))
-              subscriptionStatus <- subscriptionService
-                                     .getSubscriptionStatus(record.businessDetails.utr, record.businessDetails.postcode)
-
-              //if user is partially subscribed when coming back from gg, they now have clean creds so can complete the subscription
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/StartController.scala
```diff
@@ -76,9 +77,26 @@ class StartController @Inject()(
 
             for {
               record <- subscriptionJourneyService.getMandatoryJourneyRecord(ContinueId(continueId))
-              _ <- subscriptionJourneyService.saveJourneyRecord(
-                    record.copy(cleanCredsAuthProviderId = Some(agent.authProviderId)))
-            } yield Redirect(routes.TaskListController.showTaskList())
+              subscriptionStatus <- subscriptionService
```
stevenmallen: I wonder if we should move this whole block to the subscriptionService under a method like redirectAccordingToStatus(...) or something?  Be nice to make the controllers less complex ...

mariannepearson0: 👍

stevenmallen: Actually, the logic looks similar to handlePartiallySubscribedAndRedirect(...), no?   These two blocks do similar things, not sure how to consolidate, or even if it's worth it... thoughts?

mariannepearson0: umm i've moved it into the service. I think those 2 methods are different enough to not warrant consolidating them, there will be too many variables.

PR #433, opened 2019-09-26
[APB-3959][SM]: AA Show NI and DOB pages if they exist for an agent

sureshhmrc:


app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnector.scala
```diff
@@ -135,15 +137,14 @@ class AgentSubscriptionConnector @Inject()(
...
+    monitor(s"ConsumedAPI-Agent-Subscription-getDesignatoryDetails-GET") {
+      val url = new URL(baseUrl, s"/agent-subscription/citizen-details/${nino.value}/designatory-details").toString
       http
-        .POST[CitizenDetailsRequest, HttpResponse](new URL(baseUrl, endpoint).toString, citizenDetailsRequest)
-        .map(_.status == 200)
+        .GET[DesignatoryDetails](url)
         .recover {
-          case e: BadRequestException => false
-          case f: NotFoundException   => false
+          case f: NotFoundException =>
```
stevenmallen: I think we should at least have a Logger.warn() or Logger.error() here?  How expected is this?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/DateOfBirthController.scala
```diff
@@ -54,12 +53,22 @@ class DateOfBirthController @Inject()(
...
 
   def showDateOfBirthForm(): Action[AnyContent] = Action.async { implicit request =>
-    withSubscribingAgent { _ =>
+    withSubscribingAgent { agent =>
       withValidSession { (_, existingSession) =>
-        existingSession.dateOfBirth match {
-          case Some(dob) =>
-            Ok(html.date_of_birth(dateOfBirthForm.fill(dob)))
-          case None => Ok(html.date_of_birth(dateOfBirthForm))
+        (agent.authNino, existingSession.nino, existingSession.dateOfBirthFromCid) match {
```
stevenmallen: The logic in this method is now a little bit opaque.  The method is still called "showDateOfBirthForm", but in fact this method can now show a bunch of different things depending on state.  To make it clearer, can we either rename the method, add some scaladoc or explain the reasoning behind each case please?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/NationalInsuranceController.scala
```diff
@@ -45,15 +51,29 @@ class NationalInsuranceController @Inject()(
   def showNationalInsuranceNumberForm(): Action[AnyContent] = Action.async { implicit request =>
     withSubscribingAgent { implicit agent =>
       withValidSession { (_, existingSession) =>
-        existingSession.nino match {
-          case Some(nino) =>
-            Ok(html.national_insurance_number(ninoForm.fill(nino)))
-          case None => Ok(html.national_insurance_number(ninoForm))
+        withValidBusinessType(existingSession) {
```
stevenmallen: Again, this method, called "showNationalInsuranceNumberForm" no longer just does that - can we rename or describe the reasoning behind the new logic?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/NationalInsuranceController.scala
```diff
@@ -62,8 +82,23 @@ class NationalInsuranceController @Inject()(
           .fold(
             formWithErrors => Ok(html.national_insurance_number(formWithErrors)),
             validNino =>
-              updateSessionAndRedirect(existingSession.copy(nino = Some(validNino)))(
-                routes.DateOfBirthController.showDateOfBirthForm())
+              if (agent.authNino.flatMap(normalizeNino) == normalizeNino(validNino.value)) {
```
stevenmallen: Please move this predicate to a single line method with a name which reflects what the comparison means

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/NationalInsuranceController.scala
sureshhmrc: these show methods either just display the page or redirect to some other page  which any show page does, method names are ok I think, if we have to rename the name, we have to rename each and every point because think in terms of auth session
if session doesn't exist we then redirect to GG login , so every endpoint we have to rename  as show...OrRedirectToGGLogin ?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/VatDetailsController.scala
```diff
@@ -56,23 +57,41 @@ class VatDetailsController @Inject()(
...
+
+              case _ => Redirect(routes.BusinessTypeController.showBusinessTypeForm())
+            }
           }
         case None => Redirect(routes.BusinessTypeController.showBusinessTypeForm())
       }
     }
   }
 
+  private def withAAChecks(agent: Agent, existingSession: AgentSession)(result: Result): Result =
```
stevenmallen: This looks like a very similar structure to the "showDateOfBirthForm" code above in DateOfBirthController.  Can the two be merged into a single method that implements this logic?  Is the logic really the same?

app/uk/gov/hmrc/agentsubscriptionfrontend/models/AgentSession.scala
```diff
@@ -30,10 +30,12 @@ case class AgentSession(
   postcode: Option[Postcode] = None,
   nino: Option[Nino] = None,
   companyRegistrationNumber: Option[CompanyRegistrationNumber] = None,
-  dateOfBirth: Option[DateOfBirth] = None,
+  dateOfBirth: Option[DateOfBirth] = None, // this is user entered DOB
```
stevenmallen: I like these comments, but please use Scaladoc on the actual class :)

app/uk/gov/hmrc/agentsubscriptionfrontend/models/DesignatoryDetails.scala
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.agentsubscriptionfrontend.models
+
+import play.api.libs.json.{Format, Json}
+import uk.gov.hmrc.agentsubscriptionfrontend.models.DesignatoryDetails.Person
+
+//Add more fields as required: https://github.com/hmrc/citizen-details
```
stevenmallen: Please use scaladoc

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/NationalInsuranceController.scala
stevenmallen: Yeah, ok fair enough.  Then please add a description of the reasoning (not the what, but the why), so people understand the reason for the re-routing?  Just a little scaladoc...?

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnector.scala
sureshhmrc: I added a warning log in the postcode controller

sureshhmrc:
@stevenmallen I prefer not to add doc as they quickly become stale,  but I added to make you happy:), please review

app/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnector.scala
stevenmallen: thank you!

stevenmallen:
> @stevenmallen I prefer not to add doc as they quickly become stale, but I added to make you happy:), please review

Thanks, Suresh :)  I think we have a duty to try and keep the docs up to date (and fix where misleading), but I understand your reticence.  I think there's still a little bit of logic duplication in the code, but happy to merge this now...

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/VatDetailsController.scala
sureshhmrc: yeah, I checked the code duplication think, but I think we don't get much benefit there with a single method, I can explain why when you are in the office

PR #483, opened 2020-03-27
[APB-4798][GW] trading address: update task list

geoffreywatson:


app/uk/gov/hmrc/agentsubscriptionfrontend/models/Tasks.scala
```diff
@@ -39,29 +65,47 @@ final case class AmlsTask(isMaa: Boolean, amlsData: Option[AmlsData]) extends Ta
...
+  override val link: String = ""
+}
+
+final case class ContactTradingAddressSubTask(
+  contactDetailsTradingAddressData: Option[BusinessAddress],
+  showLink: Boolean)
+    extends SubTask {
+  override val taskKey: String = "contactDetailsTradingAddressSubTask"
+  override val isComplete: Boolean = contactDetailsTradingAddressData.isDefined
+  override val link: String = ""
```
sureshhmrc: you are overriding `link` with "" in all the subTasks, instead you can do this assignment in the base class and not override in sub classes ...
then it will be just little verbose

app/uk/gov/hmrc/agentsubscriptionfrontend/service/TaskListService.scala
```diff
@@ -35,24 +35,69 @@ class TaskListService @Inject()(agentAssuranceConnector: AgentAssuranceConnector
...
+          subscriptionJourneyRecord.contactDetailsTradingAddress,
+          contactTradingNameSubTask.isComplete
+        )
+        val contactDetailsTask: Task = ContactDetailsTask(
+          List(
+            contactEmailSubTask,
+            contactTradingNameSubTask,
+            contactTradingAddressSubTask
+          )
+        )
```
sureshhmrc: its all 20 lines of duplicated code in if and else blocks ...
can we refactor this ?

PR #484, opened 2020-03-30
APB-4800 Trading address: contact email address

geoffreywatson:


app/uk/gov/hmrc/agentsubscriptionfrontend/views/contact_email_address.scala.html
```diff
@@ -0,0 +1,72 @@
...
+)
+
+@(contactEmailForm: Form[BusinessEmail], isChange: Boolean)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+
+@backLinkUrl = @{
+    if(isChange){routes.SubscriptionController.showCheckAnswers()}else{routes.ContactDetailsController.showContactEmailCheck()}
+}
+
+@formValue = @{
+    if(contactEmailForm.data.get("email").isDefined){
```
sureshhmrc: you can use pattern matching here ,
using `.get` is not recomended

PR #507, opened 2020-09-21
GW update play-ui for accessibility statement

geoffreywatson:


conf/application.conf
```diff
@@ -219,3 +219,5 @@ mongo-async-driver {
   }
 }
 
+
+accessibility-statement.service-path = "/agent-subscription"
```
sureshhmrc, line 223: i think we dont need this as a config as its  same in different environments

PR #511, opened 2020-11-06
[APB-4951][GW] internal a11y review

geoffreywatson:


app/uk/gov/hmrc/agentsubscriptionfrontend/views/amls/amls_details.scala.html
```diff
@@ -58,6 +58,8 @@
         || error.key.contains("expiry.month") || error.key.contains("expiry.year")))
     }
 
+@value = @{ amlsForm("amlsCode").value match { case Some(x) => x case None => "" }}
```
gabor-dorcsinecz, line 61: amlsForm("amlsCode").value.getOrElse("")

PR #519, opened 2021-02-23
Apb 4996

YuanZHANG910:


app/ErrorHandler.scala
```diff
@@ -48,26 +49,34 @@ class ErrorHandler @Inject()(
...
+    }
   }
 
   override def resolveError(request: RequestHeader, exception: Throwable): Result = {
     auditServerError(request, exception)
 
     exception match {
-      case _: NoActiveSession =>
+      case e: NoActiveSession =>
+        logger.error(s"NoActiveSession ${e.getMessage}")
```
geoffreywatson, line 63: why an error here?

YuanZHANG910: That is the ticket requirement.
https://jira.tools.tax.service.gov.uk/browse/APB-4996

geoffreywatson (approved):

PR #527, opened 2021-04-19
Apb 5180 continue sub on different creds

barrysims:
- I've narrowed the type of the implicit conversion, later we should remove it.  AMLSController is a good demonstration of why we're better off without them!

- Also removed some Future[Either]s which were adding nothing

- Added OptionT to utils


geoffreywatson:
appears to be no changes to the logic - just some refactoring of the code hence no tests.

PR #531, opened 2021-04-21
[APB-5162][GW] contact details changes

geoffreywatson:


geoffreywatson:
I believe content changes were tested on previous commit

app/uk/gov/hmrc/agentsubscriptionfrontend/config/view/CYACheckResult.scala
```diff
@@ -57,7 +57,7 @@ object CYACheckResult {
     tradingNameData: ContactTradingNameData,
     tradingAddressData: ContactTradingAddressData)(reg: Registration, maybeAmls: Option[AmlsData]): CYACheckResult =
     if (emailData.contactEmail.isEmpty) FailedContactEmail
-    else if (tradingNameData.hasTradingName && tradingNameData.contactTradingName.isEmpty)
+    else if (!tradingNameData.hasTradingName && tradingNameData.contactTradingName.isEmpty)
```
barrysims, line 60: Do we need to check tradingNameData.hasTradingName here? I'm not familiar with the business logic, but it seems like the happy path could complete without checking tradingNameData.hasTradingName.

barrysims (approved):

app/uk/gov/hmrc/agentsubscriptionfrontend/config/view/CYACheckResult.scala
geoffreywatson: The logic switched over so the variable should really be called useBusinessName - but trying to avoid a breaking change by renaming it...I left a comment about it in the code under Task. Not sure of a better way to handle this.

barrysims: gotcha, thanks!

PR #537, opened 2021-08-09
[APB-5283][GW] validate HMRC amls record

geoffreywatson:


build.sbt
```diff
@@ -32,7 +32,7 @@ lazy val wartRemoverSettings = {
       Wart.ArrayEquals,
       Wart.AnyVal,
       Wart.EitherProjectionPartial,
-      Wart.Enumeration,
+      //Wart.Enumeration,
```
dg-21-hmrc, line 35: I actually agree with Wart remover in that I'm not a fan of enumerations

build.sbt
```diff
@@ -17,7 +17,7 @@ lazy val scoverageSettings = {
 lazy val wartRemoverSettings = {
   val wartRemoverWarning = {
     val warningWarts = Seq(
-      Wart.JavaSerializable,
+      //Wart.JavaSerializable,
```
dg-21-hmrc, line 20: What was the change that made disabling this check necessary?

app/uk/gov/hmrc/agentsubscriptionfrontend/models/AmlsSubscriptionRecord.scala
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package uk.gov.hmrc.agentsubscriptionfrontend.models
+
+import play.api.libs.json.{Format, Json}
+import uk.gov.hmrc.agentsubscriptionfrontend.models.FormBundleStatus.FormBundleStatus
+
+import java.time.LocalDate
+
+object FormBundleStatus extends Enumeration {
```
dg-21-hmrc, line 24: To be honest I don't like using "enumerations" very much, I much prefer plain sealed traits even at the cost of a bit more boilerplate.
Are enumerations already in wide use in this project or in related projects?

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/AMLSController.scala
```diff
@@ -182,24 +187,43 @@ class AMLSController @Inject()(
...
-                      amlsDetails = Some(AmlsDetails(supervisoryBodyData, Right(RegisteredDetails(validForm.membershipNumber, validForm.expiry))))))
-                ).map(
-                  _ => Redirect(continueOrStop(continue, routes.AMLSController.showAmlsDetailsForm()))
-                )
+                amlsService.validateAmlsSubscription(validForm).flatMap {
+                  case AmlsSuspended | _: AmlsCheckFailed => Redirect(routes.AMLSController.showAmlsRecordIneligibleStatus)
+                  case DateNotMatched | RecordNotFound    => Redirect(routes.AMLSController.showAmlsDetailsNotFound)
+                  case ResultOK => {
+                    val supervisoryBodyData =
+                      amlsBodies.getOrElse(validForm.amlsCode, throw new Exception("Invalid AMLS code"))
```
dg-21-hmrc, line 195: What does the user get if this exception is thrown?

dg-21-hmrc (commented):
I have left a few questions, mainly to help me understand

build.sbt
geoffreywatson: ok but Enumeration was a nice easy way to deal with the different statuses and they are easy to serialise

build.sbt
geoffreywatson: Enumeration json serialiser

app/uk/gov/hmrc/agentsubscriptionfrontend/models/AmlsSubscriptionRecord.scala
geoffreywatson: no not really but I don't see the problem with Enumerations...they are better than a String? which is all these Enum's represent really but happy to change it if you prefer

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/AMLSController.scala
geoffreywatson: an general error (Sorry, there is a problem with the service) but the form validator would have prevented a amls body that is not on the list.

PR #548, opened 2021-09-17
[APB-5313][GW] RAD Fix: postcode page content update

geoffreywatson:


conf/messages.cy
```diff
@@ -270,10 +270,14 @@ crn.title=Beth yw rhif cofrestru cwmni eich busnes?
 crn.hint=Er enghraifft, 01234567 neu SC123456
 
 #Postcode page
-postcode.sole_trader.title=Beth yw cod post eich cyfeiriad cofrestredig?
-postcode.limited_company.title=Beth yw cod post cyfeiriad cofrestredig eich cwmni?
-postcode.partnership.title=Beth yw cod post cyfeiriad cofrestredig eich partneriaeth?
-postcode.llp.title=Beth yw cod post cyfeiriad cofrestredig eich partneriaeth atebolrwydd cyfyngedig?
+postcode.sole_trader.title=What is the postcode of the address you registered with HMRC for Self Assessment?
```
CarolineS-QA, line 273: Do we not have Welsh yet? Or will it be in a different ticket?

geoffreywatson: we normally don't wait for Welsh to merge but we also don't deploy to prod until the Welsh is in...added on the same ticket but later commit

CarolineS-QA: I see, okay.

If you don't deploy until Welsh is in it seems weird to have English on the message keys as substitute, unless it's just previously been deployed like that when Welsh wasn't switched on - I thought it would be easier to see empty keys as 'to do' - but then you may have a page with no content sooo... fair enough?

PR #567, opened 2022-02-03
APB-5575 LLP content changes for a11y

max-hmrc:


app/uk/gov/hmrc/agentsubscriptionfrontend/views/postcode.scala.html
```diff
@@ -50,7 +50,9 @@
             labelAsHeading = true,
             labelClass = Some("govuk-label--xl"),
             hint = Some(msgs(s"postcode.${businessType.key}.hint")),
-            inputClass = Some("govuk-input--width-10")
+            hintId = Some("postcode-hint"),
+            inputClass = Some("govuk-input--width-10"),
+            id = "postcode"
```
CarolineS-QA, line 55: You shouldn't need to feed id & hintId here because govukInput uses the `.withFormField(field)` which creates the id and hintId... unless the tests say otherwise?

max-hmrc: yeah. so in the test i am testing that the hint and the label have the correct text. which is why I added them.

CarolineS-QA: that's weird though because the tests are taking the hmtl output, so I would still expect the hint and id to be rendered by the component with just the `.withFormField(field)` method - so you're saying that without this they aren't behaving like we expect from   https://github.com/hmrc/play-frontend-hmrc#withformfield  ?

it/uk/gov/hmrc/agentsubscriptionfrontend/controllers/PostcodeControllerWithAssuranceFlagISpec.scala
```diff
@@ -64,9 +73,12 @@ class PostcodeControllerWithAssuranceFlagISpec extends BaseISpec with SessionDat
...
       status(result) shouldBe 200
-      checkHtmlResultWithBodyText(result,
-        "What is the postcode of the address you registered with HMRC for Self Assessment?",
-        "This could be a home address.")
+
+      val html = Jsoup.parse(Helpers.contentAsString(Future.successful(result)))
+      html.title() shouldBe "What is the postcode of the address you registered with HMRC for Self Assessment? - " +
+        "Create an agent services account - GOV.UK"
+      html.select("label[for=postcode]").text() shouldBe "What is the postcode of the address you registered with HMRC for Self Assessment?"
+      html.select("#postcode-hint").text() shouldBe "This could be a home address."
```
max-hmrc, line 81: Here's my test @CarolineS-QA

it/uk/gov/hmrc/agentsubscriptionfrontend/support/Css.scala
```diff
@@ -0,0 +1,19 @@
...
+  val ERROR_SUMMARY_TITLE = "#error-summary-title"
+  val ERROR_SUMMARY_LINK = ".govuk-list.govuk-error-summary__list li a"
+  val errorSummaryLinkWithHref  = (href: String)  => s".govuk-list.govuk-error-summary__list li a[href=$href]"
+  val H1 = "main h1"
+  val PRE_H1 = "main .govuk-caption-l"
+  val paragraphs = "main p"
+  def errorSummaryForField(id: String): String ={
+    s".govuk-error-summary__body li a[href=#${id}]"
+  }
+  def errorForField(id: String): String = s"span#${id}-error"
```
max-hmrc, line 14: and this css selector  can be used to test that the error summary link text for a given field of a given id is there where it should be.

max-hmrc: Try now Caroline

CarolineS-QA (approved):

PR #587, opened 2022-11-04
APB-6577 Prevent user from registering if NINO is marked as deceased

dg-21-hmrc:
See APB-6577.
Currently there are two proposed versions of the new content.
Both are implemented, you can switch between the two of them through a boolean flag (BusinessIdentificationController.scala, line 452).
Once there is a decision on which version to use, please set the flag appropriately and merge.


umhmrc:
For when you are back from your break, @dg-21-hmrc : 

Do we need this PR any more now that the requirement has been overridden in https://jira.tools.tax.service.gov.uk/browse/APB-6794 ?

CarolineS-QA:
Superseded by https://github.com/hmrc/agent-subscription-frontend/pull/592

PR #607, opened 2023-07-07
[CC][APB-7268]Hacked in new page for when the dates don't match in amls

cavster:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/AMLSController.scala
```diff
@@ -234,6 +234,11 @@ class AMLSController @Inject()(
       Ok(amlsNumberNotFoundTemplate())
     }
   }
+  def showAmlsDateNotMatched: Action[AnyContent] = Action.async { implicit request =>
+    withSubscribingAgent { _ =>
+      Ok(amlsNumberNotFoundTemplate())
```
CarolineS-QA, line 239: isn't this the same view as `showAmlsNumberNotFound` ? is it meant to be the new page instead

cavster: Good spot I'll change it now

CarolineS-QA (approved):

geoffreywatson (approved):

PR #613, opened 2023-07-17
[CC]Remove lower case in getEmailToVerify to fix redirect error

cavster:


CarolineS-QA (approved):
Do you know why we were mapping to lower case in the first place? ie. are we making a case insensitive match with the email in session somewhere else

PR #615, opened 2023-08-25
[APB-7273][GW] add contact telephone number to task list

geoffreywatson:


app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
```diff
@@ -309,5 +309,33 @@ package object controllers extends Logging {
           submittedAnswer => Seq(Yes, No).contains(submittedAnswer.check)
         )
     )
+
+    val contactPhoneCheckForm = Form[ContactPhoneCheck](
```
CarolineS-QA, line 313: could we look at moving these forms from this package file into a separate 'forms' directory? just seems a bit odd compared to our other FE services

CarolineS-QA (approved):

app/uk/gov/hmrc/agentsubscriptionfrontend/controllers/package.scala
geoffreywatson: yes of course....probably a good idea. We can make a ticket for that.

PR #623, opened 2024-06-13
[WG][APB-8460]  Upgrade to play 3

gitwojciech:


build.sbt
```diff
@@ -1,3 +1,23 @@
+import uk.gov.hmrc.{DefaultBuildSettings, SbtAutoBuildPlugin}
+
+
+val appName = "agent-subscription-frontend"
+
+ThisBuild / majorVersion := 1
```
r-melvin, line 6: This is also defined on line 31, should we increment this to 2 as well?

geoffreywatson: https://semver.org/ -- this is what I was going to review with the team

geoffreywatson: although we have incremented major version in agent-assurance going to Play 3.0 I believe we shouldn't have...?

r-melvin: I'm not overly precious on the versioning as long as it makes sense, you could argue that moving to play 3 is major change as we have to change the scala version, restructure the app, fix texts and be exposed to new bugs etc

PR #630, opened 2024-09-10
APB-8688 remove support for mixed content

nubz:


nubz:
We are not able to switch into Sensitive models using encrypted values created with the methods used on this service, instead we need to maintain the encrypt/decrypt methods we started with, so closing this in favour of the alternative approach in https://github.com/hmrc/agent-subscription-frontend/compare/APB-8688-temp-remove-isEncrypted-logic?expand=1

PR #642, opened 2025-03-31
APB-9308 Message keys/Welsh translation

raw1697:


conf/messages.en
```diff
@@ -382,11 +382,11 @@ nonAgent.p1=To continue, <a href="{0}">sign in with an agent user ID</a>.
 nonAgent.p2=If you do not have one, <a href="{0}">create a new Government Gateway user ID</a>, selecting the ‘Agent’ option.
 
 #Subscription complete
-subscriptionComplete.title=Agent services account created
-subscriptionComplete.h1=Agent services account created
-subscriptionComplete.accountName=You created an account for
-subscriptionComplete.copiedAcross=You have also copied across <strong class="govuk-!-font-weight-bold">{0}</strong> clients to your new agent services account.
-subscriptionComplete.p1=Make a note of your account reference number <strong class="govuk-!-font-weight-bold">{0}</strong>. Make sure you have this with you if you contact us about your agent services account.
+subscriptionComplete.title=You’ve created an agent services account
```
nubz, line 385:
```suggestion
subscriptionComplete.title=You’ve created an agent services account
```

nubz: although this doesn't show in the GITHUB UI aand my suggestion is not usable as it detects no change (!) I have replaced the apostrophe with a smart quote - the apostrophe in the source messages file is not a smart quote and does not render in the UI (messages API does not render straight apostrophes - whilst we can escape them to render them we should be using smart quotes instead)

PR #644, opened 2025-04-23
APB-9412 created test for connector

raw1697:
Test for FE connector


it/test/uk/gov/hmrc/agentsubscriptionfrontend/connectors/AgentSubscriptionConnectorISpecIt.scala
```diff
@@ -379,6 +380,32 @@ class AgentSubscriptionConnectorISpecIt extends BaseISpecIt with MetricTestSuppo
...
+    }
+
+    " return None when agent subscription record returns a 404 response" in {
+      AgentSubscriptionStub.givenAmlsRecordNotFound("12345")
+      val result = await(connector.getAmlsSubscriptionRecord("12345"))
+
+      result shouldBe None
+    }
+
+    "throw Upstream4xxResponse if details doesn't match records" in {
```
lewishay, line 400: Would you mind just rewording this, it doesn't sound accurate with the scenario. Looking at the connector code this scenario is: throw an UpstreamErrorResponse exception when agent subscription returns an unexpected status

lewishay (changes requested):

lewishay (approved):

PR #664, opened 2026-05-21
APB-11453 Private access to legacy (pre‑MMTAR) ASA subscription journey with UTR validation against the Agent Compliance Team manually assured list

kaizar12:


conf/application.conf
```diff
@@ -135,7 +135,7 @@ tracking-consent-frontend {
 }
 
 features {
-  agent-assurance-run = false
+  agent-assurance-run = true
```
kaizar12: `Prod` has this set to `true`, should this be set to `true` by default?

VLukovski: Was probably turned off because otherwise setting up test data in our stubs is somewhat difficult

VLukovski (dismissed):

VLukovski (approved):

